PR #27, opened 2025-08-13
Planning PR 2: Save chat messages to Axon Ivy repo

mnhnam-axonivy:
Currently I have 2 problems:
1. Cannot save chat message immediately because seem that the Axon Ivy repo is asynchronized with Axon Ivy designer, java cannot find the created AgentMemory in the repo to save new message.
-> solution: while the agent is running, store messages in memory. After the agent is finished, store in the Axon Ivy repo.
image

2. The `ChatMessage` of LangChain4j cannot peristed to the Axon Ivy repo
-> solution: use `Json.toJson()` method to parse messages to a list String, then save that list instead of `ChatMessage` list.


ivy-rew:
Totally agree, let's close it for now.

PR #36, opened 2025-08-20
Change CODEOWNERS to team-wawa

mnhnam-axonivy:
Changed code owner to team-wawa

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,1 @@
-# Team Octopus is the default CODEOWNER unless you choose to take on the role
-*  @axonivy-market/team-octopus
+*  @axonivy-market/team-wawa


.github/CODEOWNERS
```diff
@@ -1,2 +1 @@
-# Team Octopus is the default CODEOWNER unless you choose to take on the role
-*  @axonivy-market/team-octopus
+*  @axonivy-market/team-wawa
```
ivy-rew, line 1: I totally support that we take ownership of this product. 
2 things:
- I'd also be codeowner here, please include me also in the wildcard selector
- In the Market I've made bad experience with generic team-responsibilities. It easily happens that nobody will actually work on the PR, since no specific person is assigned. So I'd rather add directly you as codeowner. What do you think?

ivy-rew (approved):
Great; being responsible on cool initiatives feels good :sunglasses:

PR #38, opened 2025-08-20
Allow releases of pre-liminary custom alpha/beta versions :ship: 

ivy-rew:


ivy-rew:
depends upon https://github.com/axonivy-market/github-workflows/pull/57

PR #43, opened 2025-08-21
AIAI-128 rebrand packages as 'smart-orchestrator' :magic_wand: 

ivy-rew:
step 1: rename the project coordinates, but not yet any-directories (to keep compatible with open PRs/WIP)


mnhnam-axonivy (approved):
LGTM pls merge

PR #50, opened 2025-08-25
AIAI-128 favour Agentic Process element over java-code Agents

ivy-rew:
Though I like that we can use native langChain4j code, we should favor our Agent element if possible.
- more low code ready, not forcing the user into writing java code; especially not the NEO user.
- the interface is stable and under our jurisdiction, while we have no control over langChain4j behavior nor interface design

... of course we should still opt for langChain4j native impls if the element doesn't fit. Or to demonstrate a feature which is not yet available as Program agent.


mnhnam-axonivy (approved):
LGTM please merge 👍

smart-orchestrator-demo/src/com/axonivy/utils/smart/orchestrator/demo/agent/TaskInfoAgent.java
ivy-rew, line 1: fyi @chnam-axonivy , I think you also used this class or something similiar to craft your task creation prompt.
Please adapt it as well in your demo and try to use directly the ProgramInterface instead of this.

PR #53, opened 2025-09-04
Generalize REST Response setup provisioning :bookmark_tabs: 

ivy-rew:
Introduce common test utils for frequently used resource/response setup.


smart-orchestrator-test/src_test/ch/ivyteam/test/log/ResourceResponse.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+
+import javax.ws.rs.core.Response;
+
+import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+public class ResourceResponse implements BeforeEachCallback {
```
ivy-rew, line 11: very simple generalization of our resource-loading/response pattern. 
although small, it should reduce the amount of duplicated resource-loading related code in our test base.

PR #55, opened 2025-09-05
Use smart-orchestrator REST client for integrated tracing

ivy-rew:


smart-orchestrator/src/com/axonivy/utils/smart/orchestrator/client/SmartHttpClient.java
```diff
@@ -0,0 +1,88 @@
...
+    return Optional.ofNullable(ct)
+        .map(List::getFirst)
+        .filter(String.class::isInstance)
+        .map(String.class::cast)
+        .orElse(MediaType.APPLICATION_JSON);
+  }
+
+  @Override
+  public void execute(HttpRequest request, ServerSentEventParser parser, ServerSentEventListener listener) {
+    throw new RuntimeException("not implemented");
```
ivy-rew, line 68: A shortcoming that we have to keep in mind. SSE request won't work yet ... we can implement them when needed or switch back to the normal client while it's not provided here.  
Nevertheless, the impl should be easy. I just didn't want to engineer it without having a proper test service.

PR #80, opened 2025-11-06
Drop DecisionMaker: no use-case in smart-workflow :wastebasket: 

ivy-rew:
no references in demos; nor tests ... so no longer part of the smart-workflow eco-system.
we can re-introduce it at any time if we see a benefit.

### DIFF
--- a/smart-workflow-demo/dataclasses/AgentDemo/SupportAgentData.d.json
+++ b/smart-workflow-demo/dataclasses/AgentDemo/SupportAgentData.d.json
@@ -10,9 +10,6 @@
     "name" : "query",
     "modifiers" : [ ]
   }, {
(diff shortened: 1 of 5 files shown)


ivy-rew:
As we're getting more users on this product; I think it's time to decide on the APIs which are part of it and where we invest to keep it stable. 

I think we already discussed, that the decision-maker is no longer required. But it seems there are still references in the support-agent demo. Can we drop it @mnhnam-axonivy ? 
Do you need a task or story for this?

decisions

mnhnam-axonivy:
Hi @ivy-rew ,
I also reduce some warnings

image

ivy-rew:
Thanks a lot for making this PR complete @mnhnam-axonivy :clap:

PR #82, opened 2025-11-06
ChatModel.SPI to create a Model from any vendor :birthday: 

ivy-rew:
Vision: every provider lives in another project, with it's own configuration (variables.yaml) and libraries.
- market installer UI allow to select the providers to install

Discussion:
- keep OpenAI as default impl within smart-workflow? 
  - size  
  - ease of usage/installation
  - role-model impl for next model?
- to do
  - [x] dynamic loading of all providers #83 
  - [x] test

Caveats:
- selection of libraries for providers; avoid duplication (langchain-core, http-client)

### DIFF
--- a/smart-workflow/config/variables.yaml
+++ b/smart-workflow/config/variables.yaml
@@ -11,4 +11,4 @@ Variables:
       # [enum: gpt-4o, gpt-4.1, gpt-4.1-mini, gpt-4.1-nano, gpt-5]
       Model: ""
     # [enum: OpenAI]
-    defaultProvider: ""
+    DefaultProvider: ""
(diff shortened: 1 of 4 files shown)


ivy-rew:
as the SPI loading now works; this is IMHO ready to be integrated on 'master'.
There are still a few TODOs in the code (caching), but the main functionality we need is here.

smart-workflow/config/variables.yaml
```diff
@@ -10,3 +10,5 @@ Variables:
       # See dev.langchain4j.model.openai.OpenAiChatModelName for the list of all supported models
       # [enum: gpt-4o, gpt-4.1, gpt-4.1-mini, gpt-4.1-nano, gpt-5]
       Model: ""
+    # [enum: OpenAI]
+    defaultProvider: ""
```
mnhnam-axonivy: Change to `DefaultProvider` ('d' to 'D') to be consistent with the format of other variables

PR #89, opened 2025-11-11
Move OpenAI variables; setup like any model provider

ivy-rew:
Breaks existing configs; so that we can setup OpenAI like any other vendor.


mnhnam-axonivy (approved):
LGTM please merge 👍

PR #96, opened 2025-11-14
Introduce an E2E test: calling OpenAI for structured-output :test_tube: :calling: 

ivy-rew:
As discussed; a first end2end test for OpenAI :
- by intention we fire only a single test with very deterministic results ... the goal is not to cover every corner case of the demos. But to use once the real LLM over its official endpoint, with zero mocks in charge.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,18 @@
+name: E2E-Build
+
+on:
+  schedule:
+    - cron:  '0 1 * * 0' # 1 AM on Sunday
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  checks: write
+
+jobs:
+  build:
+    uses: axonivy-market/github-workflows/.github/workflows/ci.yml@v6
+    with:
+      mvnArgs: -P e2e.test 
+    secrets:
+      mvnArgs: -DOPEN_AI_API_KEY=${{ secrets.OPEN_AI_API_KEY }}
--- a/smart-workflow-test/src_test/com/axonivy/utils/smart/workflow/model/openai/OpenAiModelIT.java
+++ b/smart-workflow-test/src_test/com/axonivy/utils/smart/workflow/model/openai/OpenAiModelIT.java
@@ -1,0 +1,47 @@
+package com.axonivy.utils.smart.workflow.model.openai;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.BeforeEach;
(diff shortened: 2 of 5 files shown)


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,18 @@
...
+  contents: read
+  checks: write
+
+jobs:
+  build:
+    uses: axonivy-market/github-workflows/.github/workflows/ci.yml@v6
+    with:
+      mvnArgs: -P e2e.test 
+    secrets:
+      OPEN_AI_API_KEY: ${{ secrets.OPEN_AI_API_KEY }}
```
ivy-rew: consume from github secrets

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,18 @@
+name: E2E-Build
+
+on:
+  schedule:
+    - cron:  '0 1 * * 0' # 1 AM on Sunday
+  workflow_dispatch:
```
ivy-rew: on request, or once a week

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,18 @@
...
+  
+permissions:
+  contents: read
+  checks: write
+
+jobs:
+  build:
+    uses: axonivy-market/github-workflows/.github/workflows/ci.yml@v6
+    with:
+      mvnArgs: -P e2e.test
```
ivy-rew: only enable e2e.test via maven profile

smart-workflow-test/src_test/com/axonivy/utils/smart/workflow/model/openai/OpenAiModelIT.java
```diff
@@ -0,0 +1,47 @@
...
+  @Test
+  void structuredOutput_e2e(BpmClient client) {
+    Ivy.session().loginSessionUser("James", "secret");
+    var res = client.start()
+        .process(AGENT_TOOLS.elementName("structuredOutput"))
+        .as().session(Ivy.session())
+        .execute();
+    TestToolUserData data = res.data().last();
+    assertThat(data.getPerson().getFirstName())
+        .isEqualTo("James");
```
ivy-rew: a test which has it's complexity, calling tools, but still should provide very deterministic results, as we enforce a structured output of person and produce the tool-value on our terms.

... I guess the same test can be run against every other model-provider with the same results :crossed_fingers:

mnhnam-axonivy (approved):
LGTM please merge 👍

PR #97, opened 2025-11-14
Reduce azure-openai library footprint :weight_lifting: 

ivy-rew:
exclude a few optional dependencies. It reduces the package size by a ~ 3 MB's

before:
deps-b4

after:
deps-after

### DIFF
--- a/smart-workflow-demo/pom.xml
+++ b/smart-workflow-demo/pom.xml
@@ -29,12 +29,6 @@
       <version>${project.version}</version>
       <type>iar</type>
     </dependency>
-    <dependency>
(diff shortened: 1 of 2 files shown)


mnhnam-axonivy:
I also simplified dependencies route:

`smart-workflow-demo`: 
- remove dependency of `smart-workflow-azure-openai` (only use by the `smart-workflow-test` project)

`smart-workflow-test`:
- remove dependency of `smart-workflow-openai` (existing dependency of `smart-workflow-demo`)
- add dependency of `smart-workflow-azure-openai`

### Before
before

### After
after

PR #98, opened 2025-11-16
Support Multi-Model: Google Gemini AI models 

mnhnam-axonivy:
Implemented same as Azure OpenAI provider


models/smart-workflow-gemini/src/com/axonivy/utils/smart/workflow/gemini/GeminiServiceConnector.java
```diff
@@ -0,0 +1,88 @@
...
+import com.axonivy.utils.smart.workflow.client.SmartHttpClientBuilderFactory;
+import com.axonivy.utils.smart.workflow.gemini.enums.GoogleAiGeminiChatModelName;
+import com.google.common.base.Objects;
+
+import ch.ivyteam.ivy.environment.Ivy;
+import dev.langchain4j.model.chat.Capability;
+import dev.langchain4j.model.googleai.GoogleAiGeminiChatModel;
+import dev.langchain4j.model.googleai.GoogleAiGeminiChatModel.GoogleAiGeminiChatModelBuilder;
+
+public class GeminiServiceConnector {
```
ivy-rew: I think this class should also reside in the model/gemini package. Maybe even in an model/gemini/internal package. I don't think users should directly program against it.

PR #101, opened 2025-11-17
Azure OpenAI as dedicated installable add-on

ivy-rew:
azure-openai as installable product, but un-selected by default ... so users have to opt-in

smart-installer

### DIFF
--- a/smart-workflow-product/product.json
+++ b/smart-workflow-product/product.json
@@ -18,6 +18,13 @@
             "version": "${version}",
             "type": "iar",
             "importInWorkspace": true
+          },
+          {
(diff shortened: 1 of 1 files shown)


smart-workflow-product/product.json
```diff
@@ -18,6 +18,13 @@
             "version": "${version}",
             "type": "iar",
             "importInWorkspace": true
+          },
+          {
+            "groupId": "com.axonivy.utils.ai",
+            "artifactId": "smart-workflow-azure-openai",
+            "version": "${version}",
+            "type": "iar",
+            "importInWorkspace": false
```
ivy-rew: opt-in

PR #165, opened 2026-02-12
AIAI-200 Image/Text/File Input for Agents

mnhnam-axonivy:
## Features
Support extracting images and PDF file in the User message:

image

Supported file types: `.png`, `.jpg`, `.jpeg`, `.pdf`
Supported model: OpenAI

## Problem

The `AiService` feature seem to be ignored `ImageContent` and `PdfFileContent` silently, therefore I use this approach:

- convert image & pdf to Base64
- For each file, use the `FileExtractor` class to make a single file extraction call to the AI provider

It works like this:

image

## Next steps

- Check real supported models of Gemini, OpenAI, Azure OpenAI and document them
- Implement guardrail for file content that should check:
  file extension
  file size (if too big it may out of LLM's window context)
  encrypted, password-protected file
  harmful files (see Portal)


smart-workflow/src/com/axonivy/utils/smart/workflow/program/internal/QueryExpander.java
```diff
@@ -0,0 +1,101 @@
...
+        Map<String, String> placeholders = new LinkedHashMap<>();
+        StringBuilder result = new StringBuilder();
+
+        while (matcher.find()) {
+            String expression = matcher.group(1).trim();
+            String uuid = UUID.randomUUID().toString();
+            placeholders.put(uuid, expression);
+            matcher.appendReplacement(result, uuid);
+        }
+        matcher.appendTail(result);
```
ivy-rew: methods that inline comments, to show sub-steps look like a smell. Please extract it into a dedicated method where the comment becomes the descriptive method name.

if you like to dive deeper on this topic; I'd strongly recommend this read:
https://www.amazon.de/-/en/Clean-Code-Handbook-Software-Craftsmanship/

smart-workflow/src/com/axonivy/utils/smart/workflow/program/internal/QueryExpander.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.regex.Pattern;
+
+import org.apache.commons.lang3.StringUtils;
+
+import com.axonivy.utils.smart.workflow.extraction.FileExtractor;
+
+import ch.ivyteam.ivy.process.program.exec.ProgramContext;
+import dev.langchain4j.model.chat.ChatModel;
+
+public class QueryExpander {
```
ivy-rew: please provide a dedicated test for this unit:
I think we will need more macro hacks in the future. 
One idea I have in mind is too get rid of standard macro expansion for complex objects. I'd rather expose them as inlined "json" instead of ivy's homebrew braces format. ... this is for later however, but I just want to make the point that this custom macro expansion is getting more important in the future

smart-workflow/src/com/axonivy/utils/smart/workflow/extraction/FileExtractor.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  public String extract(File file) {
+    if (file == null) {
+      return "";
+    }
+
+    UserMessage userMessage = UserMessage.from(createFileContent(file));
+    return model.chat(SYSTEM_PROMPT, userMessage).aiMessage().text();
+  }
```
ivy-rew: I think we should not involve the model and do special prompt-engineering for the user towards a specific use-case we gleaned from the CV-check scenario.

- too specific prompt, what could go beyond or counteracts what the user really want's to achieve
- static system prompt is may weakly compatible with other untested models
- much more testing from our side is required.
- what we inject may be the standard behaviour
- heavy token consumption

IMHO we should just send the files as is using the "Image/PdfFile-Content" payloads together with the prompt of the user. 
- easier to maintain and foresee consequences
- independent from model vendors
- in the CV-check demo we can still explicitly hint your "markdown" extraction system prompt explicitly in the demo process.

ivy-rew (commented):
Good direction 💪 . 

Let's think through the image extraction again. I think it's important to not do extra LLM calls here. I'd rather just vote to enrich the TextContent prompt with additional Image/Pdf contents. So the user has full power to steer what would should be done with the file. By example we could ask questions on the content, without actually manually expanding the content into our context.

mnhnam-axonivy:
Hi @ivy-rew ,

I just updated code of `FileExtractor` and `QueryExpander`. Now they are:

- Read `InputStream` instead of File
-  Support `Path`
- Remove the logic to re-run macro again `hasOtherExpressions`, if the object isn't file-related types, return its String representative using `String.valueOf`
- Clean code, split methods to smaller units to easier to read
- Added test for `QueryExpander`

still handling other feedback...

---

There is a problem, in the method `QueryExpander.expandFileExpressions()`, you may wonder why I use Function<String, Optional<Object>> resolver instead of passing ProgramContext directly. It was because I don't know how to mock `ProgramContext`, so I use function so in the `TestQueryExpander` I can mock result of ProgramContext.

mnhnam-axonivy:
Hi @ivy-rew 

I’ve updated the documents about the new feature:
https://github.com/axonivy-market/smart-workflow/pull/165/changes/e964608fb31e0c5407c86fb696eccbab096e8b6b

Could you please review it when you have a moment? Once approved, we can merge and release this cool feature 😄

ivy-rew:
> Hi @ivy-rew
> 
> I’ve updated the documents about the new feature: [e964608](https://github.com/axonivy-market/smart-workflow/commit/e964608fb31e0c5407c86fb696eccbab096e8b6b)
> 
> Could you please review it when you have a moment? Once approved, we can merge and release this cool feature 😄

Thanks for the update. I still feel we need to get rid of the 'model' interaction in the FileExtractor. Because it somewhat limits what can be done with files. Did you read my comments on that or have concerns?

https://github.com/axonivy-market/smart-workflow/pull/165#discussion_r2812017659

ivy-rew (approved):
Good to go as far as I see. Great job 🥇

PR #176, opened 2026-02-26
Support Binary and InputStream

mnhnam-axonivy:
This PR handle the feedback to support Binary and InputStream from the PR [Support Images and PDF Input for agent 📚 ](https://github.com/axonivy-market/smart-workflow/pull/165) 

### Implementation
- `QueryExpander` now detect Binary and convert to InputStream
- `FileExtractor`:
  - if the given InputStream has name -> get extension from the name and convert to corresponding file content.
  - otherwise -> guess extension by reading bytes

### How to detect extension of bytes:

- For png jpg: use `URLConnection.guessContentTypeFromStream()`
- For pdf: use byte matching approach - pdf files binary start with below prefix

```java
  // Signature bytes of PDF file: "%PDF-"
  private static final byte[] PDF_PREFIX = "%PDF-".getBytes(StandardCharsets.US_ASCII);
```

(Refer: [List of file signatures](https://en.wikipedia.org/wiki/List_of_file_signatures))

There are libraries such as [ApacheTikaDocumentParser](https://github.com/langchain4j/langchain4j/blob/main/document-parsers/langchain4j-document-parser-apache-tika/src/test/java/dev/langchain4j/data/document/parser/apache/tika/ApacheTikaDocumentParserTest.java) but they will make smart-workflow bigger so I simply use byte matching approach.


smart-workflow/src/com/axonivy/utils/smart/workflow/extraction/FileExtractor.java
```diff
@@ -37,13 +50,43 @@ public String extract(InputStream stream, String fileName) {
...
+      };
+    } catch (IOException e) {
+      return "";
+    }
+  }
+
+  /**
+   * PDF files cannot be reliably detected via URLConnection, so we check the prefix bytes for "%PDF-".
+   */
+  private static String resolvePdfExtension (byte[] data) {
```
ivy-rew: return `Optional<String>` ... to signal whether resolution worked or not.

ivy-rew (approved):
Very nice; I think the main thing to add is a central log in case the file can not be added as 'Content'. So that the user can ask to support his file-type or convert the file in advance on his own terms.

smart-workflow/src/com/axonivy/utils/smart/workflow/extraction/FileExtractor.java
```diff
@@ -37,13 +50,43 @@ public String extract(InputStream stream, String fileName) {
...
       case "png" -> ImageContent.from(base64, "image/png", ImageContent.DetailLevel.HIGH);
       case "jpg", "jpeg" -> ImageContent.from(base64, "image/jpeg", ImageContent.DetailLevel.HIGH);
       case "pdf" -> PdfFileContent.from(base64, "application/pdf");
       default -> null;
     };
   }
+
+  /**
+   * Detects file type based on magic bytes. Currently supports PDF, PNG, JPEG.
+   */
```
ivy-rew: I think such comments are quickly outdated and then will be missleading. While your code I simple and nice to read ... so better omit it.

ivy-rew (approved):
I think the PR contains several comments which can be considered a bad smell. 
So I'd vote to omit most of them. Or rewrite the code methods/names variables to be more expressive, so that no comments are needed.

smart-workflow/src/com/axonivy/utils/smart/workflow/extraction/FileExtractor.java
```diff
@@ -37,13 +50,43 @@ public String extract(InputStream stream, String fileName) {
   }
 
   private static Content createContent(byte[] bytes, String fileName) {
-    String extension = FilenameUtils.getExtension(fileName).toLowerCase();
     String base64 = Base64.getEncoder().encodeToString(bytes);
+    String extension = Optional.ofNullable(fileName)
+        .map(FilenameUtils::getExtension)
+        .map(String::toLowerCase)
+        .orElseGet(() -> detectExtension(bytes));
+
```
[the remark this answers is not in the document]
mnhnam-axonivy: haha sorry, just my style to remember what I did 😆 
I will remove all the comments

mnhnam-axonivy:
Hi @ivy-rew 
I updated code. Changes:
- Support new type: IDocument
- Remove unnecessary comments
- Close stream after use
- Centralize error log for exception by using `SmartWorkflowException`

Do you like the idea to use an Exception class like `SmartWorkflowException` to centralize place handle error log?

PR #198, opened 2026-03-27
XIVY-18650 Use natively supported OTLP exporter

ivy-rew:
adapt new native tracing exporter that works without instrumenting the agent: https://github.com/axonivy/core/pull/12541
see docs: https://github.com/axonivy/doc/pull/229


mnhnam-axonivy (approved):
Great feature! 🔥

PR #205, opened 2026-04-07
AIAI-388 Introduce RAG tool

mnhnam-axonivy:
Hi @ivy-rew ,

I’ve implemented RAG as a tool for OpenSearch. This is the first story to concretize the architecture and define how RAG should work in Smart Workflow.

Could you please review the approach and share your feedback? 😄

---

### OpenSearch 

I introduced new project `smart-workflow-opensearch-rag` specified for RAG using OpenSearch as the vector store.
This project expose 2 features:

### A callable process to create new vector store

image

Developer can use this callable to create vector store if it doesn't exist then ingest documents to the vector store.
Since this callable is important, I won't set it as an agent tool, I want the admin user do it manually.

### RAG tool to search the ingested documents using query

image

AI Agent can use this tool to search for documents it wants.

### New variables

Introduced new variables specified for:
- RAG: in `smart-workflow/variables/yaml`
- OpenSearch connection: in `smart-workflow-opensearch-rag/variables.yaml`

### Extend `ChatModelProvider`

From now on, the `ChatModelProvider` also contains options for embedding model and we also have `EmbeddingModelFactory` to handle embedding model creation

### RAG pipelines

Smart Workflow RAG has 2 pipelines:

- Ingest:  sources → [split] → [embed] → RagConnector → RagVectorStore.addAll()
- Query:   query   → [embed] → RagConnector → RagVectorStore.search() → RagResult

To do that, I introduced some interfaces that RAG projects could extend upon

**RagVectorStore**

The lowest layer. Wraps an actual vector database with three operations: store embeddings + text, search, and close the connection.

**RagConnector**:

Handles how to connect to a vector store for a given collection (index/namespace)

**RagIngestor**

Handles the ingestion pipeline: Splits source documents into chunks via `RagDocumentSplitter`, embeds all chunks, then stores embeddings + segments into the vector store via `RagConnector`.

**RagRetriever**

Handles the retrieval pipeline: Embeds the user's query string, builds an `EmbeddingSearchRequest` (with maxResults / minScore thresholds), then searches the vector store and maps results to RagMatch objects (text + score + metadata)

### Implement OpenSearch module

From the interfaces above, I implemented the RAG pipeline for OpenSearch in this project.

## Others

- In `doc/external-vector-stores/opensearch`, I added PowerShell and Bash scripts to quickly initialize OpenSearch vector store.
- Tests will come in a separated PR because changes in this PR is too much for a review.


doc/external-vector-stores/opensearch/start.ps1
```diff
@@ -0,0 +1,202 @@
+# WARNING: FOR DEMO/DEVELOPMENT USE ONLY — NOT FOR PRODUCTION.
+# Security is disabled. Data is unencrypted and unauthenticated.
+#
+# Starts OpenSearch via Docker Compose for use as a vector store.
+# Security plugin is disabled — HTTP only, no authentication for connections.
+# An admin password is still required by OpenSearch 2.12+ at startup.
+#
```
ivy-rew: Do we need the starts script really? 
I think its easily generated but still work to maintain and keep it sync with its SH variant.
I'd rather use static value for pass and port in the compose.yaml.
It's totally valid for demo use-cases. For production one has to set it up manually anyway.

smart-workflow/src/com/axonivy/utils/smart/workflow/rag/pipeline/RagIngestor.java
```diff
@@ -0,0 +1,35 @@
...
+import com.axonivy.utils.smart.workflow.rag.document.processor.RagDocumentSplitter;
+import com.axonivy.utils.smart.workflow.rag.entity.RagResult;
+
+import dev.langchain4j.data.embedding.Embedding;
+import dev.langchain4j.data.segment.TextSegment;
+import dev.langchain4j.model.embedding.EmbeddingModel;
+
+public interface RagIngestor {
+
+  RagConnector getConnector();
```
ivy-rew: I don't understand why pass the connector around on these interface. Seem to be rather a property we pass into the Ingestor, Retriever, Store, ... rather than something this interface controls? 🤔 

Also connection closing is implemented on all these.

ivy-rew (commented):
Cool initiative 🥇 

I think the main point to work on is the HTTP Client integration. 
We'd rather opt for a connector stack that fits into the AxonIvy Engine.
And also the external Opensearch instance from the Engine should be our first choice.

doc/external-vector-stores/opensearch/start.ps1
mnhnam-axonivy: This is a lesson learned from implementing the AI Assistant.

For presales, we needed a quick way to start the vector store, so I included a default username and password in the Docker Compose file. However, this was later reported during a bug bounty event.

So I will delete those scripts, developer need to define username/password themselves.

ivy-rew (commented):
Some more thoughts on this PR
- The libraries to connect to OpenSearch are huge: I think we should thoroughly invest for alternatives. I think in the ivy-core we also use a plain Jersey client to communicate with OpenSearch (and omitted the official library footprint also due to its size). 
- The proposed interfaces promote connection handling, but this is not necessarily needed (e.g. for EasyRAG) so we better omit it for a first version.
- We now have an 'ingest' tool. But I don't see the value here. I think it tempts the user to route Docs over an LLM and then into OpenSearch using a tool. Which costs a lot of token for a value I don't see. I'd have expected just a simple search tool. 

To address these things: I think **we should better first implemented the "EasyRAG"** variant. 
- I think it will be leaner: smaller library footprint + no connection handling. 
- It will help us to keep interfaces minimal. we can still extend or re-discuss the common interfaces once we introduce OpenSearch RAG. 

Or what's on your mind concerning a first RAG integration @mnhnam-axonivy ?

mnhnam-axonivy:
>There is no ingest tool, only a normal callable that admins can use to add documents via a process call if needed.
>Developer can use this callable to create vector store if it doesn't exist then ingest documents to the vector store.
Since this callable is important, I won't set it as an agent tool, I want the admin user do it manually.

perfect, Thanks.

---

>There are a few challenges with this approach:
    1. In-memory vector store
    
    EasyRAG uses an in-memory vector store. This means the data is lost after the process ends, making it suitable only for presales or experimentation, not for production use.
    
    2. Heavy dependencies
    
    To make RAG “easy,” the module uses heavy dependencies:
      - Apache Tika
      - A quantized embedding model (bge-small-en-v1.5). Refer: https://huggingface.co/RedHatAI/bge-small-en-v1.5-quant
    
    You can see them im the pom.xml: https://github.com/langchain4j/langchain4j/blob/main/langchain4j-easy-rag/pom.xml
    
    The build artifact becomes quite large (~174MB), which is not ideal for our use case.
    image
    
    3. The quality and cost to host an embedding model
    
    - Running an embedding model locally increases RAM usage. While it reduces external dependencies, it shifts the cost to runtime resources. I couldn’t find strong references for using bge-small-en-v1.5 in production, but it seems like a trade-off: smaller external footprint vs. higher memory consumption.
    
    - The model is quantized, meaning it is faster and smaller but less accurate. For real-world documents such as legal content or product procedures, this introduces significant risk.

Very good research. Thanks a lot @mnhnam-axonivy . So I'm definitively with you that EasyRAG is a bad choice especially for its library size. IMHO we will never need it apart from maybe testing. Go on with opensearch then :muscle: :1st_place_medal:

mnhnam-axonivy:
Hi @ivy-rew ,
I updated some code:

- Remove `DefaultCollection` variable
- Simplify the pipeline
- Adopt the Axon Ivy REST client approach. For reference, see the **smart-workflow-opensearch-rag** project, especially:
  - `OpenSearchPayloadBuilder`
  - `OpenSearchRestClient`
  - `rest-clients.yaml`
  - One point to consider:
    Instead of using a fixed authentication method defined in `rest-clients.yaml`, we could introduce an AuthFilter inside `OpenSearchRestClient`. This would allow more flexible authentication (e.g., API key or username/password) depending on the use case.

What do you think about this approach?

---

Will work on other feedback tomorrow.

rag/smart-workflow-opensearch-rag/config/rest-clients.yaml
```diff
@@ -0,0 +1,11 @@
...
+RestClients:
+  openSearch:
+    Name: openSearch
+    Url: ${ivy.var.AI.RAG.OpenSearch.Url}
+    Features:
+      - ch.ivyteam.ivy.rest.client.mapper.JsonFeature
+    Properties:
+      SensitiveData: 'true'
+      Deserialization.FAIL_ON_UNKNOWN_PROPERTIES: 'false'
+      TrustSelfSignedCertificates: ${ivy.var.AI.RAG.OpenSearch.DisableSslVerification}
```
ivy-rew: I'd opt for one term and use it as property and variable in a similar way. 
Personally I like "TrustSelfSigned.." better than the negated (DisableSSL..."

rag/smart-workflow-opensearch-rag/src/com/axonivy/utils/smart/workflow/rag/opensearch/internal/OpenSearchRestClient.java
```diff
@@ -0,0 +1,150 @@
...
+    base.register(new AuthFilter(apiKey, userName, password));
+    return new OpenSearchRestClient(base);
+  }
+
+  private OpenSearchRestClient(WebTarget target) {
+    this.target = target;
+  }
+
+  public void ping() {
+    Response response = target.request(MediaType.APPLICATION_JSON).head();
```
ivy-rew: response is AutoCloseable, please always handle resources within try-with-resources

rag/smart-workflow-opensearch-rag/src/com/axonivy/utils/smart/workflow/rag/opensearch/internal/OpenSearchRestClient.java
```diff
@@ -0,0 +1,150 @@
...
+      boolean hasApiKey = StringUtils.isNotBlank(apiKey);
+      boolean hasBasicAuth = StringUtils.isNotBlank(userName) && StringUtils.isNotBlank(password);
+
+      if (hasApiKey) {
+        addApiKeyHeader(context);
+        return;
+      }
+      if (hasBasicAuth) {
+        addBasicAuthHeader(context);
+      }
```
ivy-rew, line 195: ok for me.

rag/smart-workflow-opensearch-rag/src/com/axonivy/utils/smart/workflow/rag/opensearch/internal/OpenSearchRestClient.java
```diff
@@ -0,0 +1,150 @@
...
+
+public class OpenSearchRestClient {
+
+  private static final String OPENSEARCH_CLIENT = "openSearch";
+  private static final String NDJSON = "application/x-ndjson";
+  private static final String PATH_BULK = "_bulk";
+  private static final String PATH_SEARCH = "_search";
+  private static final int HTTP_OK = 200;
+  private static final int HTTP_SUCCESS_MAX = 300;
+  private static final int HTTP_BAD_REQUEST = 400;
```
ivy-rew: let's rather re-use JAX-RS standard status constants for such checks: javax.ws.rs.core.Response.Status.BAD_REQUEST

rag/smart-workflow-opensearch-rag/src/com/axonivy/utils/smart/workflow/rag/opensearch/internal/OpenSearchRestClient.java
```diff
@@ -0,0 +1,150 @@
...
+  private static final String ERR_BULK_PARTIAL = "OpenSearch bulk ingest had errors: %s";
+  private static final String ERR_PING_FAILED = "OpenSearch ping failed with status: %d";
+  private static final String ERR_CREATE_INDEX = "OpenSearch index creation failed for '%s' with status: %d — %s";
```
ivy-rew: minor style issue; but I'd cluster similar constants under a common parent interface.
```suggestion
private interface Error
  String BULK_PARTIAL = "OpenSearch bulk ingest had errors: %s";
  String PING_FAILED = "OpenSearch ping failed with status: %d";
  String CREATE_INDEX = "OpenSearch index creation failed for '%s' with status: %d — %s";
}  
```

Static style information (Error interface) rather than prefix matching.
Easier to comprehend and maintain.
Less verbose (private static final) is the default for interface fields.

rag/smart-workflow-opensearch-rag/src/com/axonivy/utils/smart/workflow/rag/opensearch/internal/OpenSearchRestClient.java
```diff
@@ -0,0 +1,150 @@
...
+import com.axonivy.utils.smart.workflow.rag.pipeline.internal.OpenSearchConf;
+import com.fasterxml.jackson.core.JsonProcessingException;
+
+import ch.ivyteam.ivy.environment.Ivy;
+import dev.langchain4j.data.embedding.Embedding;
+import dev.langchain4j.data.segment.TextSegment;
+import dev.langchain4j.store.embedding.EmbeddingSearchRequest;
+import dev.langchain4j.store.embedding.EmbeddingSearchResult;
+
+public class OpenSearchRestClient {
```
ivy-rew, line 26: Good, we have a similar client in our code-base, also using Jersey (-> ch.ivyteam.ivy.searchengine.client.SearchEngineClient) 
But I think it's good to build our own client here. The use cases we cover are different.

ivy-rew: We'll need this client covered by a test-case. I assume we need a real OpenSearch instance running. So I'd consider to write a @Testcontainers based test. 

by example: https://github.com/axonivy-market/mailstore-connector/blob/master/mailstore-connector-test/src_test/com/axonivy/connector/mailstore/test/MailStoreDemoTest.java

While we need opensearch rag features well tested, we should keep our global test-suite fast. But due to the involved infrastructure we will need more setup time. So I'd either include it in e2e tests or introduce a new kind in a similar manor. E.g. rag-tests?

rag/smart-workflow-opensearch-rag/src/com/axonivy/utils/smart/workflow/rag/opensearch/internal/OpenSearchPayloadBuilder.java
```diff
@@ -0,0 +1,136 @@
...
+  private static final String FIELD_K = "k";
+  private static final String FIELD_SOURCE = "_source";
+  private static final String FIELD_HITS = "hits";
+  private static final String FIELD_SCORE = "_score";
+  private static final String FIELD_ID = "_id";
+  private static final String BULK_OP_INDEX = "index";
+  private static final String FIELD_ERRORS = "errors";
+  private static final String FIELD_ERROR = "error";
+
+  private OpenSearchPayloadBuilder() {}
```
ivy-rew, line 60: adventurous: will need thorough test coverage

doc/RAG.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+### Demo
+
+The `ExternalRagDemo` process in `smart-workflow-demo` demonstrates a complete RAG pipeline:
+
+1. Loads three HR documents (insurance plans, company rules, benefits) from the project CMS.
+2. Ingests them into an OpenSearch index named `hr-knowledge` (skipped if the index already exists).
+3. Asks three HR-related questions through an AI agent that uses the `openSearchSearch` tool.
+4. Logs each answer along with the matched knowledge base segments and their similarity scores.
```
ivy-rew: ok for the first version: but I think a UI would be more helpful to demonstrate the feature
- user sends a bunch of files (or takes some from our templates) for ingest
- after the UI answers prompts from the user by searching the RAG

... this would make a very good impression on the user. It's not just some data that the RAG is processing, but your real world use-case data.

doc/RAG.md
```diff
@@ -0,0 +1,157 @@
...
+1. **Ingest** — Split your documents into chunks, generate embeddings, and store them in a vector store.
+2. **Search** — When a question arrives, embed the query, find the most similar chunks, and return them.
+3. **Answer** — The LLM receives the retrieved chunks as context and generates a grounded response.
+
+Smart Workflow provides callable subprocesses and AI tools that handle steps 1 and 2. Step 3 is handled by the `AgenticProcessCall` element, which orchestrates the LLM and tool calls automatically.
+
+## Common Configuration
+
+These Ivy variables control how documents are split and how search results are filtered. They apply to all vector store backends.
+
```
ivy-rew: I think these docs will easily be forgotten and tend to outdate. Also its sort of duplicating the things we should already have in the variables.yaml. 

So I'd omit re-stating variables. Sure it's easy to generate out of code. But today this task is not automated. Therefore I'd opt for one-source-of-truth. Once we start to work on the documentation, we can still inline a variables.yaml file as reference here.

ivy-rew (commented):
Cool progress. I'd invest in testing and reduced docs.

doc/RAG.md
mnhnam-axonivy: I implemented that feature in the old AI Assistant.
So I will:
- remove the demo from this PR
- copy the feature from AI Assistant to Smart Workflow in a new PR

mnhnam-axonivy:
Hi @ivy-rew ,

**Updated changes:**

- Remove variables from `RAG.md` since they are mentioned in variables.yaml (keep one source of truth)
- Use Jax-RS code constants instead of define new constant
- Change variable `DisableSslVerification` to `TrustSelfSignedCertificates`
- Clean code: move similar constants under a common parent interface
- Handle AutoCloseable resources within try-with-resources

**New feature:**

- Added metadata such as embedding model, chunk size, overlap size when create index. See `OpenSearchIndexMeta`

**TODO:**

I will create 2 PRs:
- demo with UI (reuse most parts from AI Assistant)
- Unit test + E2E test using test container

mnhnam-axonivy:
Hi @ivy-rew 
I merged test and the demo to this PR. Just let me know when I can merge it 😄

models/smart-workflow-openai/config/variables.yaml
```diff
@@ -11,3 +11,7 @@ Variables:
         # See dev.langchain4j.model.openai.OpenAiChatModelName for the list of all supported models
         # [enum: gpt-4o, gpt-4.1, gpt-4.1-mini, gpt-4.1-nano, gpt-5]
         DefaultModel: ""
+        # The default embedding model, if not explicitly specified.
+        #
```
ivy-rew, line 15: strange whitespace line -> 🗑️

rag/smart-workflow-opensearch-rag/config/variables.yaml
```diff
@@ -0,0 +1,23 @@
+# yaml-language-server: $schema=https://json-schema.axonivy.com/14.0-dev/config/variables.json
+# Common RAG settings (MaxResults, MinScore, ChunkSize, ChunkOverlap, EmbeddingModel.*)
+# are defined in the smart-workflow module under AI.RAG.*.
+Variables:
+  AI:
+    RAG:
+      OpenSearch:
+        # Base URL of the OpenSearch server.
+        # Demo/dev (Docker, security disabled): "http://localhost:19600"
+        # Production (AWS OpenSearch): "https://my-opensearch.us-east-1.es.amazonaws.com"
```
ivy-rew: do we have such an AWS openSearch running in the cloud already? or do you have any experience with it?
I think we should e2e test this -> just to verify that our custom http/rest-client works correctly with this provider.
I wouldn't include this in this PR -> can you create a story for this?

rag/smart-workflow-opensearch-rag/src/com/axonivy/utils/smart/workflow/rag/pipeline/internal/OpenSearchConnector.java
```diff
@@ -0,0 +1,69 @@
...
+  @Override
+  public EmbeddingModel embeddingModel() {
+    if (embeddingModel == null) {
+      embeddingModel = EmbeddingModelFactory.createFromIvyVars();
+    }
+    return embeddingModel;
+  }
+
+  @Override
+  public boolean indexExists(String collection) {
```
ivy-rew, line 38: Do we have a story for the feedback from yesterday about maintaining/updating existing indexes? I think it will get important for real world use-cases.

rag/smart-workflow-opensearch-rag/src/com/axonivy/utils/smart/workflow/rag/tool/OpenSearchRagTool.java
```diff
@@ -0,0 +1,38 @@
...
+    return "Search a pre-indexed OpenSearch knowledge base using semantic similarity.";
+  }
+
+  @Override
+  public List<ToolParameter> parameters() {
+    return List.of(
+        new ToolParameter("collection", "OpenSearch index name to query.", "java.lang.String"),
+        new ToolParameter("query", "The complete user question to use as the search query. Pass the full user question without modification.", "java.lang.String"),
+        new ToolParameter("maxResults", "Maximum number of results to return. Uses the configured default when null.", "java.lang.Integer"),
+        new ToolParameter("minScore", "Minimum similarity score threshold between 0.0 and 1.0. Uses the configured default when null.", "java.lang.Double"));
```
ivy-rew, line 27: I think these parameters are strange to expose to the LLM. Or at least I can't think of many real world use-cases where the LLM will not just figure out magic numbers.
So I think it's crucial that we can make these ToolParameters optional (not required). Therefore I suggest to add a 'required' field to the "ToolParameter" type. Default is required=true. But here we should configure it to false. ... and then we need to interprete this required flag when transforming the parameters to the schema. 
-> also no killer for this PR. We can do it a sub-sequent effort.

smart-workflow/src/com/axonivy/utils/smart/workflow/rag/entity/RagMatch.java
```diff
@@ -0,0 +1,44 @@
+package com.axonivy.utils.smart.workflow.rag.entity;
```
ivy-rew: I think RagMatch and RagResult should reside where they are referred. Having them in a separate package divides stuff that always lives together. 

```suggestion
package com.axonivy.utils.smart.workflow.rag.pipeline;
```

smart-workflow/src/com/axonivy/utils/smart/workflow/rag/entity/RagResult.java
```diff
@@ -0,0 +1,50 @@
+package com.axonivy.utils.smart.workflow.rag.entity;
```
ivy-rew:
```suggestion
package com.axonivy.utils.smart.workflow.rag.pipeline;
```

smart-workflow/src/com/axonivy/utils/smart/workflow/rag/pipeline/RagConnector.java
```diff
@@ -0,0 +1,16 @@
...
+
+import com.axonivy.utils.smart.workflow.model.EmbeddingModelFactory;
+
+import dev.langchain4j.model.embedding.EmbeddingModel;
+
+public interface RagConnector {
+
+  boolean indexExists(String collection);
+
+  RagVectorStore connect(String collection);
```
ivy-rew: 'connect' makes me think I'll have to manage an open connection afterwards. 
but I can't think of a better term. 
maybe 'open' would be better ... but its also missleading in the same direction I assume 😆

ivy-rew (approved):
Very nice 🆒. some minor suggestions and maybe follow ups to plan?

doc/external-vector-stores/opensearch/start.ps1
ivy-rew: we have a follow up -> share it as ready-to-use devcontainer ... so the bash/sh scripts are no longer required

rag/smart-workflow-opensearch-rag/config/variables.yaml
mnhnam-axonivy: Since this will require an AWS subscription, how about creating a separate story to support AWS Bedrock as well (an AI platform similar to Microsoft Azure and Google Vertex)?

This would help us become fully AWS-ready across the stack.

rag/smart-workflow-opensearch-rag/src/com/axonivy/utils/smart/workflow/rag/pipeline/internal/OpenSearchConnector.java
mnhnam-axonivy: I’ve been thinking about this quite a bit, and it feels like we’re at a crossroads.

Do we really want to take on the responsibility of managing a vector store?

In my opinion, we should avoid handling index maintenance and updates. It adds significant complexity and is better handled by specialized systems. Smart Workflow role should focus on querying existing vector stores rather than maintaining them.

In the future, we could consider introducing an official product dedicated to managing vector stores. But for now, keeping that responsibility separate might be the better approach.

What do you think?

rag/smart-workflow-opensearch-rag/config/variables.yaml
mnhnam-axonivy: Story for E2E OpenSearch on AWS
https://axon-ivy.atlassian.net/browse/AIAI-479

ivy-rew: Thanks for crafting the story. Good follow up which we can address when time allows.

rag/smart-workflow-opensearch-rag/src/com/axonivy/utils/smart/workflow/rag/pipeline/internal/OpenSearchConnector.java
ivy-rew: Thanks for sharing your thoughts. I'm totally with you. Maintaining Vector storages should not be our main capability. 
... rather we focus on querying. And just provide a simple environment to help customers setup/ingest first data. 

So I agree to not rush onto any further improvements here. Some-when we may document our thoughts on this in the user-docs however. So that users don't expect features from us we are not trying to incorporate.

rag/smart-workflow-opensearch-rag/src/com/axonivy/utils/smart/workflow/rag/tool/OpenSearchRagTool.java
mnhnam-axonivy: I created a mini story for it 😄 
https://axon-ivy.atlassian.net/browse/AIAI-480

smart-workflow/src/com/axonivy/utils/smart/workflow/rag/pipeline/RagConnector.java
mnhnam-axonivy: Because the result of this method is a vector store object, how about `getVectorStore` or `vectorStore`?

ivy-rew: Cool suggestion. I'm down for 'vectorStore' :ballot_box_with_check: 
```suggestion
  RagVectorStore vectorStore(String collection);
```

rag/smart-workflow-opensearch-rag/src/com/axonivy/utils/smart/workflow/rag/opensearch/internal/OpenSearchRestClient.java
```diff
@@ -0,0 +1,209 @@
...
+    try (Response response = target.path(indexName).path(Paths.SEARCH)
+        .request(MediaType.APPLICATION_JSON)
+        .post(Entity.json(body))) {
+      String responseBody = response.readEntity(String.class);
+      try {
+        return OpenSearchPayloadBuilder.parseSearchResponse(responseBody, request.minScore());
+      } catch (JsonProcessingException ex) {
+        throw new IllegalStateException("Failed to parse OpenSearch search response", ex);
+      }
+    }
```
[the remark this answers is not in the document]
mnhnam-axonivy: will handle in later story when we test with real OpenSearch on AWS

PR #208, opened 2026-04-07
AI Governance dashboard part 3: Agent activities inside a case

mnhnam-axonivy:


ivy-rew:
I feel the dashboard is not ready for a serious review.
- Feels widely wipe coded and therefore very hard to maintain (especially css, js,)
- If want to make this part of smart-workflow we should align on a plan how to proceed. 

I think to proceed we should start with minimal changes that already supply value:

1. small PRs; to allow serious reviewing
2. contained utility code should be tested in the same PR that introduces it.
3. a happy path web integration test should guarantee that the glued code works e2e ... this test can be very simple in the first PR - but we should not neglect testing till some-when later
4. we integrate a first PR and proceed from there.

If we are aware of half-serious efforts we went for the sake of having a great demo; it would be best to fire separate PRs for them once we have a stable basis. This will also make team discussion on it easier.

PR #211, opened 2026-04-08
AIAI-371 Standardize guardrail collection logic

lttung-axonivy:
## Pull request overview

Standardizes how guardrails are selected for agents: prefer explicitly configured guardrail names; if empty, fall back to defaults from `variables.yaml`; if still empty, apply no guardrails. The collector now resolves guardrails via SPI and applies them in the configured order.

**Changes:**
- Reworked `GuardrailCollector` to resolve guardrails by configured names (config list → variable default → none) and preserve configured order.
- Simplified agent configuration and UI listing to rely on the collector’s unified logic.
- Updated tests/docs to reference the new default-variable constants location and behavior.

### DIFF
--- a/smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/GuardrailCollector.java
+++ b/smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/GuardrailCollector.java
@@ -1,24 +1,28 @@
 package com.axonivy.utils.smart.workflow.guardrails;
 
-import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
-import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.lang3.StringUtils;
 
 import com.axonivy.utils.smart.workflow.guardrails.adapter.AbstractGuardrailAdapter;
 import com.axonivy.utils.smart.workflow.guardrails.adapter.InputGuardrailAdapter;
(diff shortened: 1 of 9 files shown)


smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/GuardrailCollector.java
```diff
@@ -1,88 +1,94 @@
...
-          .collect(Collectors.toCollection(LinkedHashSet::new)));
+    Set<String> requestedNames = new LinkedHashSet<>(
+        (filters != null && !filters.isEmpty()) ? filters : readVariableNames(defaultVariableKey));
+
+    if (requestedNames.isEmpty()) {
+      return List.of();
     }
 
-    allProviders().stream()
+    Map<String, G> guardrailsByName = allProviders().stream()
```
mnhnam-axonivy, line 76: When open the Configuration tab, method `allProviders()`  is called twice.
Pls help me improve it also. tks you 😄 

maybe a variable

```java
private final Set<GuardrailProvider> providers;
```

mnhnam-axonivy (approved):
LGTM, just need one additional improvement. See comment

smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/GuardrailCollector.java
lttung-axonivy: After discussion, keep the implement as it is, for simple, clean and easy to understand, keep collector with static methods, not use as collector object. Do not need to init object collector in Editor and reuse that object for input and output guardrail. If you think we should really reduce calling allProviders(), I could implement like that

ivy-rew (approved):

PR #216, opened 2026-04-09
Verify all ModelProviders are installable :package:

ivy-rew:
guide model provider implementors with a test veryfing the addition in product.json ... on the last review I forgot about the fact that feedback in a first review attempt ... this test preserves this requirement forever.
https://github.com/axonivy-market/smart-workflow/pull/194#issuecomment-4163172862


mnhnam-axonivy (approved):
This approach is so innovative 👏

PR #220, opened 2026-04-10
Warn on raising compiler issues :warning: 

ivy-rew:
I'd like to keep this repository free from unnoticed warnings. Especially deprecated code usages should early and automatically be reported.

This should be the solution for it:
- reports raising warning deltas by comparing to master

Vibe coded :blush: :
- will improve it as we go
- if we like it, we can contribute it to the global market ecosystem

warning-report
complain-warnings


mnhnam-axonivy (approved):
I like the idea a lot! ❤️

PR #231, opened 2026-04-13
AIAI-389 Observe guardrail execution :shield:

nhthinh-axonivy:
- Update chat history record with guardrail records
- Add listeners to record guardrail execution in chat history
- Add spans for guardrail execution in open inference tracing
- Add documentation for guardrail history recording
- Add tests for guardrail history recording


smart-workflow/src/com/axonivy/utils/smart/workflow/observability/openinference/OpenInferenceTracing.java
```diff
@@ -86,4 +106,78 @@ public void onEvent(AiServiceErrorEvent event) {
...
+          .map(r -> r.responseFromLLM())
+          .map(r -> r.aiMessage())
+          .map(AiMessage::text)
+          .orElse(null);
+      String guardrailName = event.guardrailClass().getSimpleName();
+      traceGuardrail(event, "OUTPUT", outputMessage, guardrailName);
+    }
+  }
+
+  private void traceGuardrail(GuardrailExecutedEvent<?, ?, ?> event, String type, String validatedMessage, String guardrailName) {
```
ivy-rew: Please isolate this attribute collector code in an own unit: just like ToolRecorder (in 'internal' package)

ivy-rew: please cover this attribute provisioning with a test in TestOpenInferenceSpans

smart-workflow/src/com/axonivy/utils/smart/workflow/observability/openinference/OpenInferenceTracing.java
```diff
@@ -30,12 +44,16 @@ public interface Var {
   }
 
   private final OpenInferenceCollector collector;
+  private final boolean hideInputMessages;
+  private final boolean hideOutputMessages;
```
ivy-rew: I've just generalized these options a little. So please update from master 😉

smart-workflow/src/com/axonivy/utils/smart/workflow/observability/openinference/span/GuardrailSpan.java
```diff
@@ -0,0 +1,34 @@
+package com.axonivy.utils.smart.workflow.observability.openinference.span;
+
+import java.util.List;
+import java.util.function.Supplier;
+
+import ch.ivyteam.ivy.trace.Attribute;
+import ch.ivyteam.ivy.trace.SpanInstance;
+import ch.ivyteam.ivy.trace.SpanResult;
+
+public class GuardrailSpan implements SpanInstance<Void> {
```
ivy-rew: I've generalized this span meanwhile -> please use AiSpan

smart-workflow/src/com/axonivy/utils/smart/workflow/observability/openinference/OpenInferenceTracing.java
```diff
@@ -86,4 +106,78 @@ public void onEvent(AiServiceErrorEvent event) {
...
+          .map(AiMessage::text)
+          .orElse(null);
+      String guardrailName = event.guardrailClass().getSimpleName();
+      traceGuardrail(event, "OUTPUT", outputMessage, guardrailName);
+    }
+  }
+
+  private void traceGuardrail(GuardrailExecutedEvent<?, ?, ?> event, String type, String validatedMessage, String guardrailName) {
+    Map<String, Object> attrs = new LinkedHashMap<>();
+    attrs.put(SemanticConventions.OPENINFERENCE_SPAN_KIND,
```
ivy-rew: can you share a short picture of this span in arize phoenix? I'd like to get a feeling on the attributes -> are they highly visible like in tools or llm? or are we just setting them up as raw attributes?

ivy-rew (commented):
Welcome to the smart-workflow contributors community 🤝 .

I've shortly checked the OpenInference setup and have some feedback to handle 💌

smart-workflow/src/com/axonivy/utils/smart/workflow/observability/openinference/OpenInferenceTracing.java
nhthinh-axonivy: Here are the guardrail spans in Arize Phoenix. They'll be wrapped under the AiSpan once I apply your change: 
<img width="2560" height="1307" alt="Screenshot 2026-04-13 at 16-24-51 smart-workflow - Projects" src="https://github.com/user-attachments/assets/415e4c1d-b356-4f5b-9bbf-098a843bcca7" />
<img width="2560" height="1307" alt="Screenshot 2026-04-13 at 16-24-12 smart-workflow - Projects" src="https://github.com/user-attachments/assets/62e1b2d9-ae91-4bda-8f4c-4f06061bd780" />
<img width="2560" height="1307" alt="Screenshot 2026-04-13 at 16-23-49 smart-workflow - Projects" src="https://github.com/user-attachments/assets/04b8c199-9828-4cf7-876d-118c2c571686" />
  - Input/output guardrails show up as dedicated GUARDRAIL spans, filterable via span_kind = GUARDRAIL
  - input.value / output.value render as first-class panels (like LLM spans)                                                                                                                                                                                                                                         
  - validator_name, validator_on_fail, pass/fail follow the OpenInference guardrails convention                                                                                                                                                                                                                      
  - Trace flow visible: input guardrail → LLM → tools → output guardrail     
wdyt?

nhthinh-axonivy: **Update!**
image

smart-workflow/src/com/axonivy/utils/smart/workflow/observability/openinference/OpenInferenceTracing.java
```diff
@@ -164,4 +173,42 @@ public void onEvent(ToolExecutedEvent event) {
...
 
+  private class InputGuardrailTracingListener implements InputGuardrailExecutedListener {
+
+    @Override
+    public void onEvent(InputGuardrailExecutedEvent event) {
+      String inputMessage = options.hideInput ? null : Optional.ofNullable(event.request())
+          .map(r -> r.userMessage())
+          .map(UserMessage::singleText)
+          .orElse(null);
+      String guardrailName = event.guardrailClass().getSimpleName();
```
ivy-rew, line 184: please use the original name of the delegate guardrail. And only the adapter simple name as fallback if it's not one of our guardrails.  see also -> SmartWorkflowGuardrail#name()

smart-workflow/src/com/axonivy/utils/smart/workflow/observability/openinference/OpenInferenceTracing.java
```diff
@@ -164,4 +173,42 @@ public void onEvent(ToolExecutedEvent event) {
...
+  private class OutputGuardrailTracingListener implements OutputGuardrailExecutedListener {
+
+    @Override
+    public void onEvent(OutputGuardrailExecutedEvent event) {
+      String outputMessage = options.hideOutput ? null : Optional.ofNullable(event.request())
+          .map(r -> r.responseFromLLM())
+          .map(r -> r.aiMessage())
+          .map(AiMessage::text)
+          .orElse(null);
+      String guardrailName = event.guardrailClass().getSimpleName();
```
ivy-rew, line 198: delegate name

doc/GUARDRAILS.md
```diff
@@ -129,6 +129,41 @@ public class MyGuardrailProvider implements GuardrailProvider {
 
    This SPI registration is **required** for Smart Workflow to discover and load your guardrails. The provider will be automatically loaded when agents request guardrails by name.
 
+## Guardrail Observability
```
ivy-rew, line 132: Guardrail's are one small link in the chain, that is trackable via Oberservability features. IMHO its overdone to document the details of the two implementations. 
Also I wouldn't explain the flags to enable it again. This can live in the OBSERVE.md.
So I would omit this section. Or alternatively mention guardrail tracing with a single sentence, but not go into details.
What's your opinion @mnhnam-axonivy

ivy-rew (commented):
The integration into Arize Phoenix looks great. Thanks for the update. 
If we get the real name integration I'll be very happy 😃

nhthinh-axonivy:
> The integration into Arize Phoenix looks great. Thanks for the update. If we get the real name integration I'll be very happy 😃

Yes I already tried multiple approach for it but:
 The langchain4j GuardrailExecutedEvent only exposes guardrailClass() (the Class), not the adapter instance, so the listener building the OpenInference / OpenSearch span name has no way to call delegate.getName(). Adding getName() to the interface would only affect places where we already hold the delegate (e.g. GuardrailCollector for filtering), not the observability output.                                                                                                                                                                                                         To actually surface "PromptInjectionInputGuardrail" instead of "InputGuardrailAdapter" in spans, the options are:                                                                                        
  1. Wait for langchain4j upstream — I may discuss with them and make a PR to expose the guardrail instance on the event. Then a one-line change in our listener gives us the real name.               
  2. ASM dynamic subclasses — generate one named subclass per delegate, so event.guardrailClass().getSimpleName() returns the real name. Works for built-in and custom guardrails, but adds bytecode generation (similar to StructuredOutputAgent).                                                                                                                                                           
  3. ThreadLocal in the adapter — set the name before evaluate(), read in the listener. Simple but I'd rather avoid ThreadLocals in the Ivy thread pool.                                                                                                                                                                                                  
  
I'd lean toward 1. Happy to keep the existing name() on the interface as-is for now. WDYT? @ivy-rew

ivy-rew:
> > The integration into Arize Phoenix looks great. Thanks for the update. If we get the real name integration I'll be very happy 😃
> 
> Yes I already tried multiple approach for it but: The langchain4j GuardrailExecutedEvent only exposes guardrailClass() (the Class), not the adapter instance, so the listener building the OpenInference / OpenSearch span name has no way to call delegate.getName(). Adding getName() to the interface would only affect places where we already hold the delegate (e.g. GuardrailCollector for filtering), not the observability output. To actually surface "PromptInjectionInputGuardrail" instead of "InputGuardrailAdapter" in spans, the options are:
> 
>     1. Wait for langchain4j upstream — I may discuss with them and make a PR to expose the guardrail instance on the event. Then a one-line change in our listener gives us the real name.
> 
>     2. ASM dynamic subclasses — generate one named subclass per delegate, so event.guardrailClass().getSimpleName() returns the real name. Works for built-in and custom guardrails, but adds bytecode generation (similar to StructuredOutputAgent).
> 
>     3. ThreadLocal in the adapter — set the name before evaluate(), read in the listener. Simple but I'd rather avoid ThreadLocals in the Ivy thread pool.
> 
> 
> I'd lean toward 1. Happy to keep the existing name() on the interface as-is for now. WDYT? @ivy-rew

Yes, I'm with you. Let's go for option one. In my experience the langChain4j community is very open and collaborative. Until then we keep the current approach :+1:

mnhnam-axonivy (approved):
LGTM 👍

PR #236, opened 2026-04-14
Set default temperature to 1 for all GPT-5 models

ivy-rew:
apply the only compatible gpt5 temperature also for sub-models (mini, nano)


mnhnam-axonivy:
LGTM 👍

PR #237, opened 2026-04-15
AIAI-452:  EnvironmentNotAvailable when call AgentEditor

mnhnam-axonivy:
**Problem:**
The Configuration tab cannot be loaded.

image

**Reason:**
During initialization of `SensitiveDataOutputGuardrail`, we call `Ivy.var()` to retrieve secrets. However, the Ivy environment is not accessible within the scope of `AgentEditor`, which causes the failure.

image

### DIFF
--- a/smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/output/SensitiveDataOutputGuardrail.java
+++ b/smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/output/SensitiveDataOutputGuardrail.java
@@ -36,8 +36,6 @@ public class SensitiveDataOutputGuardrail implements SmartWorkflowOutputGuardrai
         "-----BEGIN (RSA |EC |DSA |OPENSSH )?PRIVATE KEY-----");
   }
 
-  private final Set<String> variables = loadConfiguredApiKeys();
-
   @Override
   public GuardrailResult evaluate(String message) {
     if (message == null || message.isBlank()) {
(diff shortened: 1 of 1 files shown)


lttung-axonivy (approved):
Nice analysis, looks good 👌

smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/output/SensitiveDataOutputGuardrail.java
ivy-rew, line 62: This was initially a performance improvement: suggest by copilot. 
And I think copilot was right, once we do performance tests these lookup won't come for free (SPI + Var lookup is expensive). The same guardrail may be executed multiple times ... so we should do the lookup only once. So I'd change this code towards lazy initializing the "variables".

ivy-rew (approved):
Good, please try lazy initialization

mnhnam-axonivy:
Update:
To lazy load variable, I use

```java
private Set<String> variables = null;
```

Then in `containsConfiguredApiKey()` method:

```java
if (variables == null) {
  variables = loadConfiguredApiKeys();
}
```

mnhnam-axonivy:
Build failed due to some engine problem

image

ivy-rew:
> Build failed due to some engine problem

fix is soon deployed: https://github.com/axonivy/core/pull/12785

ivy-rew (approved):
Thank you 🤝

PR #242, opened 2026-04-17
refactor: remove CollectionUtils smell :construction:

ivy-rew:
I think using `CollectionUtils.isEmpty()` is bad practice and should be avoided:
- it implies we have a "null" issue with collections ... while serious code should always guarantee to produce empty collections like "List.of() or Map.of()" instead of returning null.
- extra dependency causes users to think what it does, while X.isEmpty() is clear for everyone.
- if there are actually areas where "null" is returned and its not code under our control ... we should explicitly code it to warn about that uncommon api.

### DIFF
--- a/models/smart-workflow-azure-openai/src/com/axonivy/utils/smart/workflow/model/azureopenai/internal/utils/VariableUtils.java
+++ b/models/smart-workflow-azure-openai/src/com/axonivy/utils/smart/workflow/model/azureopenai/internal/utils/VariableUtils.java
@@ -4,7 +4,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 
 import com.axonivy.utils.smart.workflow.model.azureopenai.internal.AzureOpenAiConf;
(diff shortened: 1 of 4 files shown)


models/smart-workflow-azure-openai/src/com/axonivy/utils/smart/workflow/model/azureopenai/internal/utils/VariableUtils.java
```diff
@@ -53,13 +52,9 @@ public static AzureAiDeployment getDeploymentByName(String deploymentName) {
...
     }
 
-    List<AzureAiDeployment> deployments = getDeployments();
-
-    if (CollectionUtils.isEmpty(deployments)) {
-      return null;
-    }
-
-    return deployments.stream().filter(deployment -> deploymentName.equals(deployment.getName())).findFirst()
-        .orElse(null);
```
ivy-rew, line 63: good example of a null-driven collection return statement, that causes complex code on the handler side:
- getDeployments() is code from us, and we returned "null" in an earlier phase of this class.
- now calling code gets unnecessarily complex ... sort of prevents us from using straight forwards simple stream or loop code.

mnhnam-axonivy (approved):
Thank you very much 👍

PR #254, opened 2026-04-24
AIAI-388 Introduce RAG tool: Handle copilot feedback

mnhnam-axonivy:
About this PR:
- Copilot was triggered after the `rag` branch was merged into `master`.
- I restored the `rag` branch to handle the feedback.
- However, after discussing with @lttung-axonivy , we agreed this approach could complicate the Git history.

Proposed solution:
Create a new branch `rag.copilot` to handle the Copilot feedback instead.


lttung-axonivy (approved):
Nice to improve quality 💯

PR #259, opened 2026-04-28
AIAI-321 Introduce PII guardrails 🚧

nhthinh-axonivy:


smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/entity/GuardrailResult.java
```diff
@@ -24,4 +30,12 @@ public Boolean isAllowed() {
...
   }
+
+  public boolean hasRewrite() {
+    return rewrittenMessage != null;
+  }
+
+  public String getRewrittenMessage() {
+    return rewrittenMessage;
+  }
 }
```
ivy-rew: I think consumer code get's easier if do not maintain a "hasRewrite". Rather I'd work directly with an Optional<String> that communicates this use-case.
```suggestion
  public Optional<String> getRewrittenMessage() {
    return Optional.of(rewrittenMessage);
  }
}
```

smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/pii/PiiDetector.java
```diff
@@ -0,0 +1,136 @@
...
+    for (PiiType type : PiiType.values()) {
+      Matcher matcher = type.pattern.matcher(masked);
+      StringBuffer sb = new StringBuffer();
+      while (matcher.find()) {
+        String piiValue = matcher.group();
+
+        if (type == PiiType.CREDIT_DEBIT_CARD_NUMBER && !isValidCreditCard(piiValue)) {
+          matcher.appendReplacement(sb, Matcher.quoteReplacement(piiValue));
+          continue;
+        }
```
ivy-rew: this special case signals; that we need a better abstraction around the now regex based replacements. 
I think we should define abstract replacement rules: they can be solved using regexes, but don't have to be. Or as in this case such a replacement-rule can have additional checks (isValidCreditCard). 

e.g.
```
public interface MaskingRule{
  String mask(String text, Map<String, String> replacements)
}
```

smart-workflow-demo/processes/Features/GuardrailDemo.p.json
```diff
@@ -173,6 +173,60 @@
...
+          "They do not correspond to any real persons, accounts, or systems."
+        ]
+      },
+      "connect" : [
+        { "id" : "f44", "to" : "f42" }
+      ]
+    }, {
+      "id" : "f47",
+      "type" : "ProcessAnnotation",
+      "name" : "PII in the user message is replaced with <TYPE_hash> placeholders before reaching the LLM. Covers: URL, bearer token, JWT, API key, password, AWS key, MAC, email, phone, card number, SSN, date of birth, IP. The output guardrail restores originals in the AI response.",
```
ivy-rew: I think this is the perfect place to Introduce the term "PII" before actually using it 😉

smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/input/PiiMaskingInputGuardrail.java
```diff
@@ -0,0 +1,28 @@
+package com.axonivy.utils.smart.workflow.guardrails.input;
+
+import com.axonivy.utils.smart.workflow.guardrails.entity.GuardrailResult;
+import com.axonivy.utils.smart.workflow.guardrails.entity.SmartWorkflowInputGuardrail;
+import com.axonivy.utils.smart.workflow.guardrails.pii.PiiDetector;
+import com.axonivy.utils.smart.workflow.guardrails.pii.PiiDetector.MaskingResult;
+import com.axonivy.utils.smart.workflow.guardrails.pii.PiiMaskingStore;
+
+public class PiiMaskingInputGuardrail implements SmartWorkflowInputGuardrail {
```
ivy-rew: javadoc that writes the PII term at full length would be helpful ⛑️

smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/output/PiiMaskingOutputGuardrail.java
```diff
@@ -0,0 +1,26 @@
+package com.axonivy.utils.smart.workflow.guardrails.output;
+
+import java.util.Map;
+
+import com.axonivy.utils.smart.workflow.guardrails.entity.GuardrailResult;
+import com.axonivy.utils.smart.workflow.guardrails.entity.SmartWorkflowOutputGuardrail;
+import com.axonivy.utils.smart.workflow.guardrails.pii.PiiDetector;
+import com.axonivy.utils.smart.workflow.guardrails.pii.PiiMaskingStore;
+
+public class PiiMaskingOutputGuardrail implements SmartWorkflowOutputGuardrail {
```
ivy-rew: javadoc: Pii -> personally identifiable information

smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/pii/PiiDetector.java
```diff
@@ -0,0 +1,136 @@
...
+
+public class PiiDetector {
+
+  public record MaskingResult(String maskedText, Map<String, String> placeholderToOriginal) {
+    public boolean hasPii() {
+      return !placeholderToOriginal.isEmpty();
+    }
+  }
+
+  private enum PiiType {
```
ivy-rew: The PiiType -> Rules should be held in a dedicated registry ... apart from the actual masking/restoring logic. So it becomes very easy to extend/adapt for third-parties.

smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/pii/PiiMaskingStore.java
```diff
@@ -0,0 +1,63 @@
...
+  private static final ConcurrentHashMap<String, Entry> STORE = new ConcurrentHashMap<>();
+  private static final AtomicInteger putCount = new AtomicInteger();
+
+  record Entry(Map<String, String> placeholderToOriginal, long createdMs) {}
+
+  static long ttlMs() {
+    try {
+      String value = Ivy.var().get(TTL_VARIABLE);
+      if (value != null && !value.isBlank()) {
+        return Long.parseLong(value.strip()) * 60 * 1000L;
```
ivy-rew: TimeUnit.Minutes.toMillis

smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/pii/PiiMaskingStore.java
```diff
@@ -0,0 +1,63 @@
...
+      }
+    } catch (Exception ignored) {}
+    return DEFAULT_TTL_MINUTES * 60 * 1000L;
```
ivy-rew: TimeUnit

smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/pii/PiiMaskingStore.java
ivy-rew: long resolution should be part of our common utility -> IvyVar

smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/pii/PiiMaskingStore.java
```diff
@@ -0,0 +1,63 @@
+package com.axonivy.utils.smart.workflow.guardrails.pii;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import ch.ivyteam.ivy.environment.Ivy;
+
+public class PiiMaskingStore {
```
ivy-rew: I think this store is hard to maintain and to do right. I think we should rather just have a guardrail which implements input and output interface at the same time. ... holding the state of the current invocation as field. Then the information will live as long as the agent and its guardrails do. Alternatively we can also share a state object between the two, since we are providing the factory the instantiates both variants.

here's a similar discussion around the same topic.
https://github.com/langchain4j/langchain4j/pull/4604#issuecomment-4019076361

smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/provider/DefaultGuardrailProvider.java
```diff
@@ -4,17 +4,19 @@
...
+import com.axonivy.utils.smart.workflow.guardrails.input.PiiMaskingInputGuardrail;
 import com.axonivy.utils.smart.workflow.guardrails.input.PromptInjectionInputGuardrail;
+import com.axonivy.utils.smart.workflow.guardrails.output.PiiMaskingOutputGuardrail;
 import com.axonivy.utils.smart.workflow.guardrails.output.SensitiveDataOutputGuardrail;
 
 public class DefaultGuardrailProvider implements GuardrailProvider {
   @Override
   public List<SmartWorkflowInputGuardrail> getInputGuardrails() {
-    return List.of(new PromptInjectionInputGuardrail());
+    return List.of(new PromptInjectionInputGuardrail(), new PiiMaskingInputGuardrail());
```
ivy-rew: If we like this guardrail; I think we should consolidate the SecretInject guardrail. IMHO these guardrails are sort of preventing the same stuff. While this one seems to run a more mature approach

ivy-rew (commented):
Cool idea: a few technical improvements to apply.

smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/pii/PiiMaskingStore.java
ivy-rew: I'm aware that we need a persist/restore mechanism if this guardrail state should survive human-in-the-loop interactions. I think we can re-use the memory store mechanism once its on master. So we don't have to maintain multiple stores around the same topic. To start with I think this guardrail can be implemented without supporting HITL.

smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/provider/DefaultGuardrailProvider.java
nhthinh-axonivy: My current understanding is that they cover different threat models. 
PII masking is a round-trip mask/restore for the LLM, while SensitiveDataOutputGuardrail blocks sensitive data from leaking in the response. Should I merge them or just make sure they don't duplicate logic?

nhthinh-axonivy:
> Cool idea: a few technical improvements to apply.

Thanks for the thorough review! 
Agreed on all points.

smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/provider/DefaultGuardrailProvider.java
nhthinh-axonivy: I also removed URL, JWT, bearer token, API key, password, and AWS key from the PII registry, those are access secrets, not personally identifiable information. Mixing them in PiiMaskingGuardrail was a scope creep.
PII now covers: email, phone, credit card, SSN, date of birth, IP address, MAC address.
In follow up maybe we can create a ticket for CredentialMaskingGuardrail, that's also where overlap with SensitiveDataOutputGuardrail can be reviewed.

ivy-rew: Thanks for the clarification. I like the separation of the scope. So I think we don't need a follow up anymore now where we have clear responsibilities

chnam-axonivy (approved):

PR #262, opened 2026-04-30
XIVY-18649 Serializer for ChatMessages :mailbox_with_mail: 

ivy-rew:
Ground work :construction: : persistent storage of the context for Human-in-the-Loop requests.

~~A lot of Jackson magic is required to get ChatMessages serialized. By doing it we can be absolutely certain that we can persist and load back all messages of the current memory.~~

~~Unfortunately there's no standard API available for that in the LangChain4j core. Translation from generic messages to serializable objects is a responsibility of the ChatModelProvider. 
See for instance  `dev.langchain4j.model.openai.internal.OpenAiUtils#toOpenAiMessages`.~~

~~By now I think serializing on our terms is the better choice:
:ballot_box_with_check: smaller compiled file and leaner dependencies in our main (smart-workflow.iar)
:ballot_box_with_check: agnostic: no specific provider knowledge in our core
:warning: tradeoff: we need to keep the API stable and follow LangChain4j changes~~


ivy-rew:
Ok: just discovered the native ChatMessageSerializer/Deserializer which I overlooked in my first attempt :pensive:

PR #265, opened 2026-05-04
MARP-4232 Update readme

ntqdinh-axonivy:
Using:

- skillset of generate-ivy-readme
- model: gpt-5-mini
- reasoning level: high
- consume: 0 premium request

### DIFF
--- a/smart-workflow-product/README.md
+++ b/smart-workflow-product/README.md
@@ -2,6 +2,8 @@
 
 **Smart Workflow** brings AI directly into Axon Ivy, so developers can build, run, and improve AI agents inside existing Axon processes. It lets business workflows leverage large language models to understand natural language, make autonomous decisions, and adapt to changing requirements — all without heavy architectural changes.
 
+### Key features
+
 Key benefits of Smart Workflow:
 
 - **Familiar setup:** Drop AI agents into BPMN processes with no structural changes and configure everything through Axon Ivy’s standard interfaces.
@@ -424,6 +426,8 @@ To function effectively, AI agents require tools to perform tasks. Smart Workflo
 
 For step-by-step instructions on creating both tool types, see the [Tools Guide](../doc/TOOLS.md).
 
+![Tool configurations screenshot showing available tools and settings](img/tool-configurations.png)
+
 ### Defining AI agent
 
 To define an AI agent, create a program element backed by the `com.axonivy.utils.smart.workflow.AgenticProcessCall` Java bean. In the `Configuration` tab, you can access and customize detailed settings for your AI agent.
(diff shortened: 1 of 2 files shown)


smart-workflow-product/README.md
```diff
@@ -463,3 +467,103 @@ To address this need, the Smart Workflow AI agent can produce output as a Java o
 You can easily configure this by specifying both the expected result type and the target object to map the result to in the `Output` section.
 
 ![Other configurations](img/agent-other-configurations.png)
+
+## Components
+
+### Exposed CALLABLE_SUB processes
+
+##### `invokeAgent(String,String,List<String>,Class)`
```
ivy-rew: this is an internal hook for the Portal. I think we don't want to expose it to other public use-cases. Have you considered only listing tagged sub processes? IMHO only "connector" tagged sub processes should be listed.

smart-workflow-product/README.md
```diff
@@ -463,3 +467,103 @@ To address this need, the Smart Workflow AI agent can produce output as a Java o
...
+
+### Maven artifacts
+
+1. com.axonivy.utils.ai:smart-workflow-demo
+
+```xml
+<dependency>
+  <groupId>com.axonivy.utils.ai</groupId>
+  <artifactId>smart-workflow-demo</artifactId>
+  <version>${version}</version>
```
ivy-rew: I'd omit the version everywhere ... or are you planning to interpolate it on the website with a real latest version?

ivy-rew (changes requested):

smart-workflow-product/README.md
ntqdinh-axonivy: I suppose that the callable sub listing mentioned that only subs with "connector" tag is listed but it was not strictly followed .  I would update the checking section for it.

smart-workflow-product/README.md
ntqdinh-axonivy: Initially, We intended to replace it with specific version as what we did in maven tab in marketplace:
https://market.axonivy.com/smart-workflow?version=14.0.0-b4#dependency

But would you like to display version in Component tab, or we could omit it completely?

ivy-rew: I think we can pick whatever works best and adjust if a customer is asking for it.
For me personally omitting the version completely is fine. And keeping a dynamic version which is replaced at runtime is equally ok. So you may pick on your preference.

smart-workflow-product/README.md
```diff
@@ -463,3 +467,103 @@ To address this need, the Smart Workflow AI agent can produce output as a Java o
 You can easily configure this by specifying both the expected result type and the target object to map the result to in the `Output` section.
 
 ![Other configurations](img/agent-other-configurations.png)
+
+## Components
+
```
ivy-sgi, line 81: This is nice 👍 
Q: Shouldn't reusable UI components be listed here as well? Are they available in Smart Workflow?

ntqdinh-axonivy: I checked and found UI components is only provided in demos project (currently, we will not mention UI dialogs/ components from demo or test project) -> there is no info about UI Components listed in this readme.

Please help me to confirm if we need to change that logic 😅

mnhnam: I confirmed, there is no UI component in Smart Workflow 😄

smart-workflow-product/README.md
```diff
@@ -1,569 +1,214 @@
...
+- **Namespace:** Portal
+- **Component type:** Data Class
+- **Fields:**
+   - `query` (String) — The user query/message to send to the AI agent
+   - `systemMessage` (String) — System message to define agent behavior
+   - `tools` (List<String>) — List of tool names available to the agent. Keep empty to use all tools
+   - `resultType` (Class) — Expected result type class for structured output. Keep null for String result
+   - `resultObject` (Object) — The AI agent response result
+- **Where used:** `SmartWorkflowAgent.p.json` (invokeAgent CallSubStart)
+- **Purpose:** Provides parameters and result mapping for invoking the Smart Workflow Agent from Portal processes
```
ivy-rew: why should we promote our internal data? the interface of the callable sub is the only public artifact.

smart-workflow-product/README.md
```diff
@@ -1,569 +1,214 @@
...
-Demo flow (start **Shopping Store** process)
-
-1. Shopper: types or says “I need a $100 red dress for a party tonight.”
-2. `Product agent` extracts attributes and expands the query (synonyms, acceptable price range: $80–$120).
-3. Axon Ivy Business Data turns criteria into an optimized filters and search for the products.
-4. Return the top products matched criteria.
+Check the demo implementations provided in the `smart-workflow-demo` module, which include user-facing examples and integrations.
 
-To quickly set up the demo data, start **Create data for shopping demo** from the process list.
+### Demo workflows
```
ivy-rew: I think we aligned to start "only" with the "components" section ... but your PR still looks as if we are trying to achieve everything with a big bang solution. please reduce the scope if you are looking for feedback

PR #267, opened 2026-05-07
Dedicated naming pattern for End-2-End tests

ivy-rew:
- clarify supported test-scenarios with explicit names that can be referred to in docs/discussions


mnhnam-axonivy (approved):
LGTM 👍

PR #279, opened 2026-05-22
Introduce ArchUnit tests to enforce module encapsulation 🥡️

ivy-rew:
Enforce architecture guidelines;
- no leacking of internals to consuming projects
- no using of "model" internas in consuming projects

Up till now I tried to stick to these rules in reviews, 
now we have it as code that fails and guides change contributors. :fast_forward: ... faster feedback loop than human reviews :1st_place_medal:

### DIFF
--- a/smart-workflow-test/pom.xml
+++ b/smart-workflow-test/pom.xml
@@ -27,6 +27,12 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>com.tngtech.archunit</groupId>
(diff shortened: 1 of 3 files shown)


ivy-rew:
Here's an example of a failing test: https://github.com/axonivy-market/smart-workflow/runs/77352065212 
I think very insightful and easy to fix when happening.

archUnit-violationReport

mnhnam-axonivy (approved):
This enforced approach is eye-opening to me. 👍

PR #281, opened 2026-05-22
Fix: OpenSearchRag can not be installed form the Market 📦️

ivy-rew:
OpenSearchRag was not selectable to be installed from the Market-installer dialog:
workflow-installer-rag-missing

... in most cases it was provided anyway, but only by "accident" through the smart-workflow-demo dependency to it.

### DIFF
--- a/smart-workflow-test/src_test/com/axonivy/utils/smart/workflow/rag/TestRagProviders.java
+++ b/smart-workflow-test/src_test/com/axonivy/utils/smart/workflow/rag/TestRagProviders.java
@@ -1,0 +1,39 @@
+package com.axonivy.utils.smart.workflow.rag;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.stream.Stream;
+
+import org.apache.commons.lang3.StringUtils;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import com.axonivy.utils.smart.workflow.market.ProductJson;
+import com.axonivy.utils.smart.workflow.spi.internal.SpiLoader;
+import com.axonivy.utils.smart.workflow.tools.provider.SmartWorkflowToolsProvider;
+
+import ch.ivyteam.ivy.environment.IvyTest;
+
+@IvyTest
+class TestRagProviders {
+
+  @ParameterizedTest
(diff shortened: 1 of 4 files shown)


smart-workflow-test/src_test/com/axonivy/utils/smart/workflow/rag/TestRagProviders.java
```diff
@@ -0,0 +1,39 @@
...
+class TestRagProviders {
+
+  @ParameterizedTest
+  @MethodSource("ragProviders")
+  void installer_promotesProvider(SmartWorkflowToolsProvider provider) throws Exception {
+    var artifactIds = ProductJson.installerArtifactIds();
+    assertThat(artifactIds)
+        .as("RAG provider is installable as extra market product")
+        .contains(installerName(provider));
+  }
```
ivy-rew, line 27: a first test that verifies, that this will never happen again for upcoming other RAG providers (if any)

smart-workflow-test/src_test/com/axonivy/utils/smart/workflow/rag/TestRagProviders.java
```diff
@@ -0,0 +1,39 @@
...
+  private static String installerName(SmartWorkflowToolsProvider provider) throws Exception {
+    var vendor = StringUtils.substringBefore(provider.getClass().getSimpleName(), "RagToolProvider") ;
+    return "smart-workflow-"+vendor.toLowerCase()+"-rag";
+  }
+
+  static Stream<SmartWorkflowToolsProvider> ragProviders() {
+    var classLoader = TestRagProviders.class.getClassLoader();
+    return SpiLoader.findImpl(SmartWorkflowToolsProvider.class, classLoader).stream()
+        .filter(p -> p.getClass().getPackageName().contains(".rag."));
+  }
```
ivy-rew, line 38: works for now; but I think for the future we'd better introduce an explicit SPI to load RAG Providers. 
... they may are incorporated as tool as we do today. But this shouldn't be the only option. 
... at least via code it should be possible to also provide a classic RAG setup, where inference is triggered by the agent framework directly (not via tool)

however, out of scope for this PR, now we are happy that we have a test at all on the market rag integration.

mnhnam-axonivy (approved):
very nice! please merge 👍

PR #293, opened 2026-06-02
Azure OpenAI provider use HTTP client 🏋🏻‍♀️

nhthinh-axonivy:
Routes Azure-OpenAI traffic through the Ivy Jersey REST client, try to close #120.

### DIFF
--- a/models/smart-workflow-azure-openai/pom.xml
+++ b/models/smart-workflow-azure-openai/pom.xml
@@ -44,6 +44,10 @@
           <groupId>dev.langchain4j</groupId>
           <artifactId>langchain4j-core</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>com.azure</groupId>
+          <artifactId>azure-core-http-netty</artifactId>
+        </exclusion>
       </exclusions>
     </dependency>
     <dependency>
--- a/models/smart-workflow-azure-openai/pom.xml
+++ b/models/smart-workflow-azure-openai/pom.xml
@@ -44,10 +44,6 @@
           <groupId>dev.langchain4j</groupId>
           <artifactId>langchain4j-core</artifactId>
         </exclusion>
-        <exclusion>
-          <groupId>com.azure</groupId>
-          <artifactId>azure-core-http-netty</artifactId>
-        </exclusion>
       </exclusions>
     </dependency>
     <dependency>
--- a/models/smart-workflow-azure-openai/pom.xml
+++ b/models/smart-workflow-azure-openai/pom.xml
@@ -44,6 +44,42 @@
           <groupId>dev.langchain4j</groupId>
           <artifactId>langchain4j-core</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>io.netty</groupId>
+          <artifactId>netty-tcnative-boringssl-static</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>io.netty</groupId>
+          <artifactId>netty-tcnative-classes</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>io.netty</groupId>
+          <artifactId>netty-transport-native-epoll</artifactId>
+        </exclusion>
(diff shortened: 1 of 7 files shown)


ivy-rew:
This looks great; did you check whether we can drop dependencies due to our own HTTP-client impl?
This would make the azure model provider much smaller and leaner to use.
http-client-azure

nhthinh-axonivy:
> This looks great; did you check whether we can drop dependencies due to our own HTTP-client impl? This would make the azure model provider much smaller and leaner to use. http-client-azure

  Yes, really good catch! I excluded **azure-core-http-netty** which drops the bulk of the Netty stack: reactor-netty-http, native transports (epoll/kqueue), and BoringSSL.
  A smaller **netty-codec-http2** sub-tree stays because langchain4j-azure-open-ai itself has hard class-level references to Netty:
   - AzureOpenAiExceptionMapper catches io.netty.channel.ConnectTimeoutException and InternalAzureOpenAiHelper
     references NettyAsyncHttpClientProvider as a default fallback. 

That part can't be excluded.

nhthinh-axonivy:
> > This looks great; did you check whether we can drop dependencies due to our own HTTP-client impl? This would make the azure model provider much smaller and leaner to use. http-client-azure
> 
> Yes, really good catch! I excluded **azure-core-http-netty** which drops the bulk of the Netty stack: reactor-netty-http, native transports (epoll/kqueue), and BoringSSL. A smaller **netty-codec-http2** sub-tree stays because langchain4j-azure-open-ai itself has hard class-level references to Netty:
> 
>     * AzureOpenAiExceptionMapper catches io.netty.channel.ConnectTimeoutException and InternalAzureOpenAiHelper
>       references NettyAsyncHttpClientProvider as a default fallback.
> 
> 
> That part can't be excluded.

  Actually had to revert that. InternalAzureOpenAiHelper has a reference to NettyAsyncHttpClientProvider as a default fallback, so the JVM throws NoClassDefFoundError even when our provider is used. The class has to stay on the classpath. 
  Would need a langchain4j change (e.g. ServiceLoader lookup instead of a direct class ref) to make this possible.

ivy-rew:
> > > This looks great; did you check whether we can drop dependencies due to our own HTTP-client impl? This would make the azure model provider much smaller and leaner to use. http-client-azure
> > 
> > 
> > Yes, really good catch! I excluded **azure-core-http-netty** which drops the bulk of the Netty stack: reactor-netty-http, native transports (epoll/kqueue), and BoringSSL. A smaller **netty-codec-http2** sub-tree stays because langchain4j-azure-open-ai itself has hard class-level references to Netty:
> > ```
> > * AzureOpenAiExceptionMapper catches io.netty.channel.ConnectTimeoutException and InternalAzureOpenAiHelper
> >   references NettyAsyncHttpClientProvider as a default fallback.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That part can't be excluded.
> 
> Actually had to revert that. InternalAzureOpenAiHelper has a reference to NettyAsyncHttpClientProvider as a default fallback, so the JVM throws NoClassDefFoundError even when our provider is used. The class has to stay on the classpath. Would need a langchain4j change (e.g. ServiceLoader lookup instead of a direct class ref) to make this possible.

Too bad. But thanks for trying it out :+1: .

I think just excluding the native libraries could be an option that works here. Netty isn't too bad, but these native impls are really huge and adding a big dependency payload.

native-libs

nhthinh-axonivy:
Done! Excluded the native platform libs, it now significantly lighter.
Filed an issue with langchain4j to fix this via ServiceLoader so when done it should be slim fit 😁

ivy-rew:
Thanks a lot @nhthinh-axonivy , your contribution is highly welcome :handshake: . 
I'll mention the issue here for ease of tracking: https://github.com/langchain4j/langchain4j/issues/5352

ivy-rew (approved):

PR #296, opened 2026-06-03
APO demo part 3: Initialize Supplier Onboarding demo

mnhnam-axonivy:
- Initialize entities classes
- Initialize the SupplierOnboarding process - the main demo process
- Added all CMS files (because it's so hard to split CMS)
- Added all CSS files (because it's so hard to split CSS)
- Implemented Supplier Request form


smart-workflow-demo/src/com/axonivy/utils/smart/workflow/demo/erp/supplier/model/SupplierCertification.java
```diff
@@ -0,0 +1,74 @@
+package com.axonivy.utils.smart.workflow.demo.erp.supplier.model;
+
+import com.axonivy.utils.smart.workflow.demo.erp.document.LegalDocumentType;
+import dev.langchain4j.model.output.structured.Description;
+
+public class SupplierCertification {
```
ivy-rew, line 6: I think keeping the model in plain java is somewhat not natural for an ivyproject. 
Contents seems to be classic javabeans, so I see no reason for the java usage.
If we can I'd switch to DataClasses's instead.

smart-workflow-demo/src/com/axonivy/utils/smart/workflow/demo/erp/supplier/onboarding/AgentProcessingStep.java
```diff
@@ -0,0 +1,187 @@
+package com.axonivy.utils.smart.workflow.demo.erp.supplier.onboarding;
+
+import java.io.Serializable;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.List;
+
+import dev.langchain4j.model.output.structured.Description;
```
ivy-rew: wouldn't JsonPropertyDescription() from jackson also work? 
If we can we should avoid using langChain4j api directly in the demo. Everything built with smart-workflow should remain vendor agnostic.

if we lack this now, we should invest into it .... I remember that we support it for tool parameters.

smart-workflow-demo/src/com/axonivy/utils/smart/workflow/demo/erp/supplier/onboarding/FindingSeverity.java
```diff
@@ -0,0 +1,29 @@
+package com.axonivy.utils.smart.workflow.demo.erp.supplier.onboarding;
+
+import dev.langchain4j.model.output.structured.Description;
+
+@Description("Severity level of a validation finding")
+public enum FindingSeverity {
+  PASSED ("so-finding-green",  "ti-circle-check",  "so-badge-green",  "so-log-line-ok",      0),
+  WARNING("so-finding-yellow", "ti-alert-triangle", "so-badge-yellow", "so-log-line-warning", 1),
+  FAILURE("so-finding-red",    "ti-circle-x",       "so-badge-red",    "so-log-line-error",   2);
```
ivy-rew: looks like a mix of severity information form the java-model with CSS UI rendering. I think the UI information should live somewhere else.

smart-workflow-demo/src/com/axonivy/utils/smart/workflow/demo/erp/supplier/onboarding/OnboardingRequest.java
```diff
@@ -0,0 +1,241 @@
...
+
+  public void setAuditTrail(List<AuditTrailEntry> auditTrail) {
+    this.auditTrail = auditTrail;
+  }
+
+  /**
+   * Builds a structured list of label-value summary lines from this request's fields.
+   * Used to populate a REQUEST-type {@link AuditTrailEntry} at submission time.
+   */
+  public List<RequestSummaryLine> buildSummaryLines() {
```
ivy-rew: I'd test it with a unit-test ...  so you see the rendered output at a glance.

smart-workflow-demo/src_hd/com/axonivy/utils/smart/workflow/demo/erp/supplier/onboarding/components/SupplierOnboardingSteps/SupplierOnboardingSteps.xhtml
```diff
@@ -0,0 +1,109 @@
...
+    </c:when>
+    <c:when test="#{s == 'SUPPLIER_DATA'}">
+      <c:set var="currentStep" value="3" />
+    </c:when>
+    <c:when test="#{s == 'DB_CHECK'}">
+      <c:set var="currentStep" value="2" />
+    </c:when>
+    <c:otherwise>
+      <c:set var="currentStep" value="1" />
+    </c:otherwise>
```
ivy-rew, line 32: can't we move this logic into a bean? logic in a view is hard to maintain.

smart-workflow-demo/src/com/axonivy/utils/smart/workflow/demo/erp/supplier/onboarding/AuditTrailEntry.java
```diff
@@ -0,0 +1,218 @@
...
+  public String getBubbleClass() {
+    if (entryType == AuditEntryType.APPROVAL) return "so-tl-bubble-completed";
+    return actorType != null ? actorType.bubbleClass : "so-tl-bubble-user";
+  }
+
+  /** Tabler icon name for the timeline bubble — APPROVAL entries use user-check. */
+  public String getIcon() {
+    if (entryType == AuditEntryType.APPROVAL) return "ti-user-check";
+    return actorType != null ? actorType.icon : "ti-user";
+  }
```
ivy-rew: I'd keep UI rendering separate from my models.

smart-workflow-demo/webContent/resources/css/supplier-onboarding.css
ivy-rew, line 1: I'd consider to move these from a naming rule into a direct sub-directory -> resource/css/supplier
The demo is that huge, that it could easily live in its totally dedicated ivy project. If we stick to clear isolation, it may be easier to move it later on.

What do you think about a dedicated demo project? I think it would prevent accidental re-usage of stuff that was designed for the supplier demo. Also it will force us to think about components which we need to generalize and share between demos or even as general component for all smart-workflow users.

ivy-rew (approved):
Good - quickly scrolled it and did a high level review; 
I think a dedicated demo project is worth to be considered to ease maintenance.

smart-workflow-demo/webContent/resources/css/supplier-onboarding.css
mnhnam-axonivy: Thank you, great idea.

I also think so, now we have features demos mixed with business demos and in upcoming sprints we will have components demos.

I think we could move business demos to that new project:
- Support Chatbot
- Shopping Demo
- Supplier Onboarding (new)
- Procurement Request (new)

I think something like "smart-workflow-business-demo" or "smart-workflow-usecase-demo", do you have idea for project name?

ivy-rew: A: good minimal choice to start would IMHO be to introduce a new directory like 'rag' and 'models', now containing 'demos'. I'd prefer 'usecase' over 'business'. 
```
|- demos
    |-- smart-workflow-demo (containing basic feature demonstrations and patterns)
    |-- smart-workflow-usecase-demo
```

---

B: Dedicated directory for every bigger use-case.
This would enable users to actually copy and customize from our demos without disentangling our demo.
```
|- demos
    |-- smart-workflow-demo (containing basic feature demonstrations and patterns)
    |-- smart-workflow-chatbot-demo
    |-- smart-workflow-shopping-demo
    |-- smart-workflow-onboarding-demo
    |-- smart-workflow-procurement-demo
```

---

C: mix of these, only separate large use-cases to own projects. But keep the generic 'usecase' as container for smaller things.

```
|- demos
    |-- smart-workflow-demo (containing basic feature demonstrations and patterns)
    |-- smart-workflow-usecase-demo
    |-- smart-workflow-onboarding-demo
```

---

I think any of these 3 choices makes our demos easier to maintain and access to the user.
I'd prefer B or C. But basically we can delay a final decision and start to isolate first self-sufficient demos.

mnhnam-axonivy: I prefer option C.

One concern is that the `procurement` demo and the `onboarding` demo currently share the same chatbot component. Could we give each project its own chatbot component? We can then refactor them to use the official chatbot component once it becomes available.

mnhnam-axonivy: so I think next steps we should:
- release new version without the "APO demo" PRs
- Handle feedback and merge current "APO demo" PRs to master
- mark the APO demo story as resolved
- create a new story to refactor demos structure
- create PRs on the new demo projects so you can continue to review

what do you think?

ivy-rew: Sorry for the late feedback; I had a day off :smile: 

> One concern is that the `procurement` demo and the `onboarding` demo currently share the same chatbot component. Could we give each project its own chatbot component? We can then refactor them to use the official chatbot component once it becomes available.

That's a positive finding. Improving our demo-code organization shapes the development of the re-usable components which are helpful also for other third-parties. I think this makes crafting the chat-component a first class priority task :arrow_up: 


> so I think next steps we should:
>     * release new version without the "APO demo" PRs
>     * Handle feedback and merge current "APO demo" PRs to master
>     * mark the APO demo story as resolved
>     * create a new story to refactor demos structure
>     * create PRs on the new demo projects so you can continue to review

Good plan, let's make it.

PR #297, opened 2026-06-05
Fix: OpenInference multimodal spans and replace null output with tool call names

### LINKED ISSUE #292: OpenInference: image/vision agent calls produce no LLM span (singleText() throws in RequestRecorder)
## Summary
When an agent (`AgenticProcessCall`) sends a **multimodal / image** user message, **no OpenInference (Phoenix) LLM span is produced**. The underlying HTTP call still shows up in the engine-cockpit REST trace, so the call "works", but it is invisible to OpenInference observability. Text-only agent calls in the same run trace normally.

## Root cause
`RequestRecorder.convertMessages()` builds the `USER` message content with `UserMessage.singleText()`:

`smart-workflow/src/com/axonivy/utils/smart/workflow/observability/openinference/internal/RequestRecorder.java` (USER branch in `convertMessages`)

`dev.langchain4j.data.message.UserMessage.singleText()` throws a `RuntimeException` when the message is **not** a single text content — i.e. when it contains `ImageContent` (image, or image + text).

`inputMessages()` wraps this in a `try/catch` that only catches `JsonProcessingException`, so the runtime exception from `singleText()` propagates out. In `OpenInferenceTracing.RequestListener.onEvent()` the order is:

```java
collector.onRequest(event.request());   // throws here for an image request
llmSpan = Span.open().instance(...);     // never reached -> no LLM span
```

So the span is never opened, while the actual model HTTP request still goes out (hence the cockpit trace).

## Reproduction
1. Set `AI.Observability.Openinference.Enabled = true`.
2. Run the **Agent Pipeline** demo (`smart-workflow-demo/processes/Patterns/AgentPipeline.p.json`) — step **"Extract Invoice Content from Image"** sends an invoice image.
3. Observe: the call appears in the engine-cockpit REST trace but produces **no** LLM span in OpenInference/Phoenix. Text-only agents in the same run produce spans normally.

## Impact
All vision / multimodal agent calls are silently missing from OpenInference observability.

## Proposed fix
1. In `RequestRecorder` stop using `UserMessage.singleText()` for the `USER` case; handle multimodal content safely (concatenate `TextContent` parts, represent non-text contents e.g. as `[image]`).
2. Broaden the `catch` in `inputMessages()` from `JsonProcessingException` to `Exception`/`RuntimeException` so a recorder failure degrades gracefully (span still created with reduced detail) instead of aborting span creation.
3. Add a test: an agent request containing an `ImageContent` user message still produces an OpenInference span.

nhthinh-axonivy:
Close #292 
- fix multimodal spans to include tool call names instead of null output
- fixed OpenInference crashes on media input

### DIFF
--- a/smart-workflow/src/com/axonivy/utils/smart/workflow/observability/openinference/internal/OpenInferenceCollector.java
+++ b/smart-workflow/src/com/axonivy/utils/smart/workflow/observability/openinference/internal/OpenInferenceCollector.java
@@ -13,6 +13,8 @@ import com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility;
 import com.fasterxml.jackson.annotation.PropertyAccessor;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
+import dev.langchain4j.agent.tool.ToolExecutionRequest;
+import dev.langchain4j.data.message.AiMessage;
 import dev.langchain4j.data.message.TextContent;
 import dev.langchain4j.data.message.UserMessage;
 import dev.langchain4j.model.chat.request.ChatRequest;
@@ -73,6 +75,18 @@ public class OpenInferenceCollector {
     return content.getClass().getSimpleName()
         .replaceAll("(?i)(File)?Content$", "")
         .toLowerCase();
+  }
+
+  public static String resolveContent(AiMessage aiMessage) {
+    if (aiMessage.text() != null) {
+      return aiMessage.text();
+    }
+    if (aiMessage.hasToolExecutionRequests()) {
+      return aiMessage.toolExecutionRequests().stream()
+          .map(ToolExecutionRequest::name)
+          .collect(Collectors.joining(", ", "[tool calls: ", "]"));
+    }
+    return null;
   }
 
   public void onRequest(ChatRequest request) {
--- a/smart-workflow/src/com/axonivy/utils/smart/workflow/observability/openinference/internal/OpenInferenceCollector.java
+++ b/smart-workflow/src/com/axonivy/utils/smart/workflow/observability/openinference/internal/OpenInferenceCollector.java
@@ -4,6 +4,7 @@ import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Objects;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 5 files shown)


smart-workflow/src/com/axonivy/utils/smart/workflow/observability/openinference/internal/OpenInferenceCollector.java
```diff
@@ -60,6 +65,30 @@ public Map<String, Object> getAttributes() {
...
+
+  public static String resolveContent(AiMessage aiMessage) {
+    if (aiMessage.text() != null) {
+      return aiMessage.text();
+    }
+    if (aiMessage.hasToolExecutionRequests()) {
+      return aiMessage.toolExecutionRequests().stream()
+          .map(ToolExecutionRequest::name)
+          .collect(Collectors.joining(", ", "[tool calls: ", "]"));
+    }
```
ivy-rew, line 88: I'm ok to render this.
Though I'm not sure whether it adds relevant data. As we already have other "attributes" to record the tool invocation request.

ivy-rew (approved):
Thanks a lot @nhthinh-axonivy . I love your pro-active resolutions 🔥

smart-workflow/src/com/axonivy/utils/smart/workflow/observability/openinference/internal/OpenInferenceCollector.java
nhthinh-axonivy: You're totally right, the tool span and the detailed tool_calls attributes are already there. The idea I want to bring up here is: it's the summary field rendered in the span list view, visible at a glance before you expand any attributes. When the LLM:

-  Responds with tool calls instead of text, output.value was null, which makes the span look like the model returned nothing, a developer scanning the trace might think something went wrong.
image

-  The intent of [tool calls: extractHeaderInfo, extractLineItems] in output.value isn't to duplicate the attribute data, it's to answer the question "what did this LLM turn do?" at the summary level, the same way a text response fills that field. The full detail is still in the attributes and the child tool spans exactly as before.
image
image

ivy-rew: I see yes it's indeed helpful. Thanks for the clarification + screenshots :heavy_check_mark:

PR #302, opened 2026-06-11
Refactor demo part 2: Data model and domain foundation (Java)

mnhnam-axonivy:
- Data model and domain foundation
- Update pom and product.json


pom.xml
```diff
@@ -15,7 +15,7 @@
     <module>models/smart-workflow-xai</module>
     <module>models/smart-workflow-anthropic</module>
     <module>rag/smart-workflow-opensearch-rag</module>
-    <module>smart-workflow-demo</module>
+    <module>demo/smart-workflow-supplier-demo</module>
```
ivy-rew, line 18: good to add the smart-workflow-supplier-demo: but don't remove the existing demo project which still exists
```suggestion
    <module>demo/smart-workflow-supplier-demo</module>
    <module>smart-workflow-demo</module>
```

demo/smart-workflow-supplier-demo/src/com/axonivy/utils/smart/workflow/demo/utils/FileProcessingUtils.java
```diff
@@ -0,0 +1,42 @@
...
+
+import ch.ivyteam.ivy.cm.ContentObject;
+import ch.ivyteam.ivy.environment.Ivy;
+import ch.ivyteam.ivy.scripting.objects.Binary;
+
+public class FileProcessingUtils {
+
+  private static final String FAILED_TO_LOAD_CMS_FILE_MSG = "Failed to load file from CMS: %s";
+
+  public static InputStream loadInputStreamFromCms(String cmsPath) {
```
ivy-rew: seems to be worth a simple unit test.

demo/smart-workflow-supplier-demo/src/com/axonivy/utils/smart/workflow/demo/utils/IvyAdapterService.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.function.Predicate;
+
+import org.apache.commons.collections4.CollectionUtils;
+
+import ch.ivyteam.ivy.process.call.SubProcessCallStartEvent;
+import ch.ivyteam.ivy.process.call.SubProcessSearchFilter;
+import ch.ivyteam.ivy.process.call.SubProcessSearchFilter.SearchScope;
+import ch.ivyteam.ivy.security.exec.Sudo;
+
+public class IvyAdapterService {
```
ivy-rew: will we really need the full blown IvyAdapterService ?
I'd assume that normal SubProcessCall will be sufficient? 
I think this will be clearer if we only add it with the change that actually consumes this.
And if possible reduce it to the minimum API actually required.

demo/smart-workflow-supplier-demo/src/com/axonivy/utils/smart/workflow/demo/utils/VectorStoreRestClient.java
```diff
@@ -0,0 +1,121 @@
...
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.client.WebTarget;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+import org.apache.commons.lang3.StringUtils;
+
+import ch.ivyteam.ivy.environment.Ivy;
+
+public class VectorStoreRestClient {
```
ivy-rew: Looks to me like an utility that should rather be shared from the rag impl? Or at least parts of it. 
Please add it when this code will be in use /consumed.

demo/smart-workflow-supplier-demo/src/com/axonivy/utils/smart/workflow/demo/utils/VectorStoreRestClient.java
```diff
@@ -0,0 +1,121 @@
...
+            if (next == 'n')      { sb.append('\n'); pos += 2; continue; }
+            if (next == 't')      { sb.append('\t'); pos += 2; continue; }
+            if (next == '\\')     { sb.append('\\'); pos += 2; continue; }
+          }
+          sb.append(c);
+          pos++;
+        }
+        texts.add(sb.toString());
+      }
+      return texts;
```
ivy-rew: text collection looks adventurous: should be encapsulated in a dedicated method describing what we're doing. Also It should be covered with a test.

ivy-rew (approved):
Adding models looks good.
Please skip the utilities that are not yet consumed nor tested.

demo/smart-workflow-supplier-demo/src/com/axonivy/utils/smart/workflow/demo/utils/VectorStoreRestClient.java
mnhnam-axonivy: removed

demo/smart-workflow-supplier-demo/src/com/axonivy/utils/smart/workflow/demo/utils/VectorStoreRestClient.java
mnhnam-axonivy: removed. I will find more simple way to normalize text when added this file again

demo/smart-workflow-supplier-demo/src/com/axonivy/utils/smart/workflow/demo/utils/IvyAdapterService.java
mnhnam-axonivy: Adapted to call SubProcess via Java
see method `AssistantUploadSupport.startSubProcessInSecurityContext()`

PR #303, opened 2026-06-11
Refactor demo part 3: Supplier business logic & services (Java)

mnhnam-axonivy:
All service, builder, helper, repository Java classes + document/assistant/employee/department infrastructure.


demo/smart-workflow-supplier-demo/src/com/axonivy/utils/smart/workflow/demo/assistant/UploadedDocumentEntryFactory.java
```diff
@@ -0,0 +1,31 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.Locale;
+
+public class UploadedDocumentEntryFactory {
+
+  private UploadedDocumentEntryFactory() {}
+
+  public static UploadedDocumentEntry of(String fileName, byte[] data) {
```
ivy-rew, line 12: this UploadedDocumentEntry may should better remain a java model? So that you can implement the utility function directly on it. Rather than in this helper

demo/smart-workflow-supplier-demo/src/com/axonivy/utils/smart/workflow/demo/assistant/AssistantUploadSupport.java
```diff
@@ -0,0 +1,316 @@
...
+  String CMS_NO_FILE_UPLOADED = CMS_SAA + "NoFileUploadedMessage";
+  String CMS_INVALID_FILE     = CMS_SAA + "InvalidFileMessage";
+  String CMS_FILE_EMPTY       = CMS_SAA + "FileEmptyMessage";
+  String CMS_FILES_QUEUED_TPL = CMS_SAA + "FilesQueuedTemplate";
+  String CMS_FILES_READY_TPL  = CMS_SAA + "FilesReadyTemplate";
+  String CMS_FILE_READY       = CMS_SAA + "FileUploadedReadyMessage";
+  String CMS_CONFIRM_PARSE    = CMS_SAA + "ConfirmParseRequired";
+  String CMS_NO_VALUES        = CMS_SAA + "NoValuesExtracted";
+  String CMS_PARSE_SUCCESS    = CMS_SAA + "ParseSuccess";
+  String CMS_PARSE_FAILED_TPL = CMS_SAA + "ParseFailedTemplate";
```
ivy-rew: wrap similar constants behind a common interface ... rather than creating a long list of constants for multiple target.

demo/smart-workflow-supplier-demo/src/com/axonivy/utils/smart/workflow/demo/document/LegalDocumentBuilder.java
```diff
@@ -0,0 +1,28 @@
...
+
+  public LegalDocumentBuilder objectId(String v) { doc.setObjectId(v); return this; }
+  public LegalDocumentBuilder objectType(LegalDocumentObjectType v) { doc.setObjectType(v); return this; }
+  public LegalDocumentBuilder documentType(LegalDocumentType v) { doc.setDocumentType(v); return this; }
+  public LegalDocumentBuilder fileName(String v) { doc.setFileName(v); return this; }
+  public LegalDocumentBuilder contentType(String v) { doc.setContentType(v); return this; }
+  public LegalDocumentBuilder fileContent(byte[] v) { doc.setFileContent(v); return this; }
+  public LegalDocumentBuilder fileSize(long v) { doc.setFileSize(v); return this; }
+  public LegalDocumentBuilder description(String v) { doc.setDescription(v); return this; }
+  public LegalDocumentBuilder uploadedNow() {
```
ivy-rew, line 20: Looks to me as if we're storing files in business data. Why are we doing this? 
- I think files from the user should be handled by ivy document storage.
- If we also need the contents for RAG processing, ok ... but this should be a duplicate which we re-create from the files whenever needed. 
- If think in Document centric BusinessData we may store objects that are closely related to files ... but the file itself should only be a path or an id which we can use to actually load the file from document storage.

ivy-rew (changes requested):
A lot of java service code and utilities ...  I think this code deserves a lot of unit-tests which make the intentions and expectations clear. I think we should have a dedicated test-project for this demo.

demo/smart-workflow-supplier-demo/src/com/axonivy/utils/smart/workflow/demo/document/LegalDocumentBuilder.java
mnhnam-axonivy: I will refactor to use process data instead, don't need to persist anything.
Thank you for pointed it out

mnhnam-axonivy:
Hi @ivy-rew 
About this feedback

> A lot of java service code and utilities ... I think this code deserves a lot of unit-tests which make the intentions and expectations clear. I think we should have a dedicated test-project for this demo.

Should I create a project `smart-workflow-supplier-demo-test` in the same demo package?

ivy-rew:
> Hi @ivy-rew About this feedback
> 
> > A lot of java service code and utilities ... I think this code deserves a lot of unit-tests which make the intentions and expectations clear. I think we should have a dedicated test-project for this demo.
> 
> Should I create a project `smart-workflow-supplier-demo-test` in the same demo package?

Yes I think creating a dedicated test project as sibling would be the best fit. 
Regarding test-contents I'd vote to at least cover the services with happy path tests. 
It's ok for me if these are more documenting general behavior, rather than being fine grained unit-tests which test valid/invalid/and corner case scenarios.
At any rate we need something that allows further development and changes, without a coding agent or manual efforts.

demo/smart-workflow-supplier-demo/src/com/axonivy/utils/smart/workflow/demo/document/RequiredDocumentUploader.java
```diff
@@ -0,0 +1,74 @@
...
+public interface RequiredDocumentUploader extends DocumentUploader {
+
+  Set<LegalDocumentType> MANAGED_TYPES = Set.of(
+      LegalDocumentType.COMMERCIAL_REGISTER,
+      LegalDocumentType.SELF_DECLARATION,
+      LegalDocumentType.ANNUAL_REPORT,
+      LegalDocumentType.CERTIFICATION
+  );
+
+  default void uploadRequiredDocument(FileUploadEvent event) {
```
ivy-rew, line 21: I think this method has some not obvious switches. So I'd cover it with a test: https://github.com/axonivy-market/smart-workflow/pull/304/changes 
Also I'd assume that by testing, we also change some of the interfaces ... since they were hard to test. Here the FileUploadEvent from PF is very prominent. Maybe we can provide some of the logic without it and therefore test it easier?

demo/smart-workflow-supplier-demo/src/com/axonivy/utils/smart/workflow/demo/document/RequiredDocumentUploader.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    LegalDocument doc = LegalDocumentBuilder.builder()
+        .objectId(ensureObjectId())
+        .objectType(getObjectType())
+        .documentType(docType)
+        .fileName(event.getFile().getFileName())
+        .contentType(event.getFile().getContentType())
+        .fileContent(event.getFile().getContent())
+        .fileSize(event.getFile().getSize())
```
ivy-rew: the LogcalDocument seems widely driven by what the file upload event of Primefaces offers. Do we really need the file size here? Does the contentType help us or can we live good enough with just the file name (and its extension)?

demo/smart-workflow-supplier-demo/src/com/axonivy/utils/smart/workflow/demo/document/RequiredDocumentUploader.java
mnhnam-axonivy: I fixed this problem in the branch `demo,supplier.test` because need to implement test cases for it.

Commit: https://github.com/axonivy-market/smart-workflow/pull/304/changes/0b37862b738862063b34d1495e24b752a59d012a

solution: 

logic put in `default void uploadRequiredDocument(String fileName, String contentType, byte[] fileContent, long fileSize)` and the `FileUploadEvent` in a wrapper method

demo/smart-workflow-supplier-demo/src/com/axonivy/utils/smart/workflow/demo/document/RequiredDocumentUploader.java
mnhnam-axonivy: Fixed in commit : https://github.com/axonivy-market/smart-workflow/pull/304/changes/d70248749589cd69d1fe1e94b47b465ff38e47d2

ivy-rew (approved):
Looks good

PR #308, opened 2026-06-15
Create a dedicated pipeline to run demo tests

mnhnam-axonivy:
- Create workflow for demo test
- Document what to do when create demo for Smart Workflow


ivy-rew (approved):
Very good; I'love small and fast pipelines with a clear scope: Thanks 🤝

pom.xml
```diff
@@ -49,4 +48,13 @@
     </pluginManagement>
   </build>
 
+  <profiles>
+    <profile>
+      <id>demo.test</id>
+      <modules>
+        <module>demo/smart-workflow-supplier-demo-test</module>
```
ivy-rew, line 55: I think we should also skip the other tests from being executed.
There's no benefit from re-running the other CI tests in smart-workflow. 
The pipeline will run faster. And we will not run into the action test-amount restriction again.

mnhnam-axonivy: Updated.

```
mvnArgs: -pl demo/smart-workflow-supplier-demo-test -am -P demo.test
```

PR #317, opened 2026-06-22
Part 8.5: Fix test process config, add sudo, and cover duplicate-check audit capture

mnhnam-axonivy:
- Fix test process configuration (ivy.case.uuid(), null params, etc.)
- Add sudo to repository access
- Add duplicateCheck_capturesAuditEntry test with mock


ivy-rew (approved):
Thanks 👍

PR #324, opened 2026-06-25
fix: skip JSON schema format for Ollama when tools are present

nhthinh-axonivy:
- Ollama does not support `format: {json_schema}` and `tools` in the same request.
- Fix current Ollama E2E test failing because model cannot pickup `tool`.


ivy-rew (approved):
Looks good, Thanks for your pro-active engagement 🤝

PR #325, opened 2026-06-25
feat: make classifier system prompt configurable for AiPromptInjectionInputGuardrail

nhthinh-axonivy:
- Add variable `classifierSystemPrompt` to `AiPromptInjectionInputGuardrail` to allow customization of the system prompt used by the classifier.


ivy-rew (approved):
Very nice addition. And very handy to do demos ... not compile/deploy round-trip needed.

PR #329, opened 2026-07-01
Governance Part 2: Implement History Table

mnhnam-axonivy:
- Initialize Governance process
- Implement UI for History table with 3 levels: Case > Task > Agent
- Implement filter capabilities for History table

image


ivy-rew (commented):
I think the GovernanceDashboard should live in its own project, as it drives the focus of being a core-library away into the UI sphere. 
What do you think about 'smart-workflow-governance' as new project?

ivy-rew:
It's a web UI, we should early on verify the functionality with a web-test. Let's introduce a dedicated test-project for this, as it will need more dependencies and another build-lifecycle.

mnhnam-axonivy:
> It's a web UI, we should early on verify the functionality with a web-test. Let's introduce a dedicated test-project for this, as it will need more dependencies and another build-lifecycle.

Which name do you prefer: `smart-workflow-webtest` or `smart-workflow-ui-test`?

ivy-rew:
> > It's a web UI, we should early on verify the functionality with a web-test. Let's introduce a dedicated test-project for this, as it will need more dependencies and another build-lifecycle.
> 
> Which name do you prefer: `smart-workflow-webtest` or `smart-workflow-ui-test`?

I prefer the '-webtest' variant :wink:

models/smart-workflow-openai/src/com/axonivy/utils/smart/workflow/model/openai/OpenAiModelProvider.java
```diff
@@ -39,7 +39,7 @@ public ChatModel setup(ModelOptions options) {
   @Override
   public List<String> models() {
     return Stream.of(OpenAiChatModelName.values())
-        .map(OpenAiChatModelName::name)
+        .map(OpenAiChatModelName::toString)
```
ivy-rew, line 42: why this change? seem like a weaker strategy than using the 'name'.

smart-workflow-test/src_test/com/axonivy/utils/smart/workflow/model/TestChatModelProviders.java
```diff
@@ -1,11 +1,10 @@
 package com.axonivy.utils.smart.workflow.model;
 
-import static org.assertj.core.api.Assertions.assertThat;
-
 import java.util.List;
 import java.util.Locale;
 import java.util.stream.Stream;
 
+import static org.assertj.core.api.Assertions.assertThat;
```
ivy-rew: ?

smart-workflow-webtest/src_test/com/axonivy/utils/smart/workflow/governance/webtest/GovernanceDashboardTest.java
```diff
@@ -0,0 +1,129 @@
...
+import com.codeborne.selenide.SelenideElement;
+import static com.codeborne.selenide.CollectionCondition.size;
+import static com.codeborne.selenide.CollectionCondition.sizeGreaterThan;
+import static com.codeborne.selenide.Condition.exist;
+import static com.codeborne.selenide.Condition.text;
+import static com.codeborne.selenide.Condition.visible;
+import static com.codeborne.selenide.Selenide.$$;
+
+@IvyWebTest(browser="chrome")
+class GovernanceDashboardTest implements LoginFixture, GovernanceDashboardFixture {
```
ivy-rew: I think this test needs a custom pipeline to run in. So that the CI remains fast and stable. Also I'd postfix it with "*WebTest" as name convention.

ivy-rew:
> I think the GovernanceDashboard should live in its own project, as it drives the focus of being a core-library away into the UI sphere. What do you think about 'smart-workflow-governance' as new project?

any concerns about this point @mnhnam-axonivy ? for me it's essential that we keep UI components out of the smart-workflow core. This makes maintenance and review much easier. 

Like it is now I think it's very tempting to add UI features unto core components. But the general history recording should not directly know anything about it's dashboard and reporting infrastructure. By separating these concepts in the code with a project, we can manifest this rule best.

demo/smart-workflow-supplier-demo/src_hd/com/axonivy/utils/ai/components/SingleLegalDocument/SingleLegalDocumentProcess.p.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema" : "https://json-schema.axonivy.com/14.0-dev/project/process.json",
-  "id" : "19FF001122330008",
+  "id" : "19FF001122330009",
```
ivy-rew: intended or accident?

ivy-rew (changes requested):
The dashboard is cool, but it should not enforce the changes it does now in 'smart-workflow'. Processes, CMS, src_Hd, webcontent ... it sounds all like a UI initiative which is easier to drive forward in an own project.

demo/smart-workflow-supplier-demo/src_hd/com/axonivy/utils/ai/components/SingleLegalDocument/SingleLegalDocumentProcess.p.json
mnhnam-axonivy: It was intentional. While reviewing the build logs, I noticed a few minor errors and initially thought I could fix them in the same PR as well.

However, I changed my mind. Here's a separate PR that contains the trivial changes and error fixes:
https://github.com/axonivy-market/smart-workflow/pull/339

mnhnam-axonivy:
> I think the GovernanceDashboard should live in its own project, as it drives the focus of being a core-library away into the UI sphere. What do you think about 'smart-workflow-governance' as new project?

I agree.
I handled and answer questions from this PR. Pls check and after that I will create the project.
And about name: which is the better name `smart-workflow-governance` or `smart-workflow-dashboard` ?

models/smart-workflow-openai/src/com/axonivy/utils/smart/workflow/model/openai/OpenAiModelProvider.java
mnhnam-axonivy: This change was initially triggered by a UI bug where the filter dropdown displayed the enum name instead of the user-friendly display name.

image

While investigating, I noticed that this issue only occurred with the OpenAI provider. After comparing the implementations, I found that the other providers use `toString()` to return the display name, whereas the OpenAI provider did not.

```java
GoogleAiGeminiChatModelName::toString
AnthropicChatModelName::toString
```

To make the behavior consistent across all providers and fix the UI issue, I updated the OpenAI provider to use toString() as well.

smart-workflow-test/src_test/com/axonivy/utils/smart/workflow/model/TestChatModelProviders.java
mnhnam-axonivy: YAGNI, I reverted

ivy-rew:
> > I think the GovernanceDashboard should live in its own project, as it drives the focus of being a core-library away into the UI sphere. What do you think about 'smart-workflow-governance' as new project?
> 
> I agree. I handled and answer questions from this PR. Pls check and after that I will create the project. And about name: which is the better name `smart-workflow-governance` or `smart-workflow-dashboard` ?

I'd love to go with 'smart-workflow-dashboard' . It gives us more flexibility to pack general adminstration views into it.

models/smart-workflow-openai/src/com/axonivy/utils/smart/workflow/model/openai/OpenAiModelProvider.java
ivy-rew: ok, I understand this is more streamlined now. I'm not sure if we are doing us a favour by using the display-name as configuration value. Maybe the #name would be more stable? But we don't know yet. So I'm fine with streamlining for now.

PR #334, opened 2026-07-06
SpiLoader.load() returns duplicate provider instances when multiple PMVs share the same SPI registration

mnhnam-axonivy:
### Root cause

`pmvsInScope()` iterates the main PMV plus all dependent PMVs. For each PMV, `findImpl()` instantiates providers via reflection (newInstance()), producing distinct objects per call. The previous deduplication used distinct() and HashSet, both of which rely on object identity — so two instances of the same class were treated as different elements.

### Fix

Replace distinct() + Collectors.toSet() with  a manual HashMap + putIfAbsent loop.
This ensures each provider class appears exactly once regardless of how many PMVs declare it.

### How to reproduce

- revert code of `SpiLoader`
- Run the new test case `TestSpiLoader.load_noDuplicates_samePmvTwice`, you will see duplication error.

### Bonus

reduce VS Code warnings in `SpiLoader`

image

### DIFF
--- a/smart-workflow-test/src_test/com/axonivy/utils/smart/workflow/model/spi/internal/TestSpiLoader.java
+++ b/smart-workflow-test/src_test/com/axonivy/utils/smart/workflow/model/spi/internal/TestSpiLoader.java
@@ -35,7 +35,7 @@ class TestSpiLoader {
     var loader = new SpiLoader(pmv) {
       @Override
       protected Stream<IProcessModelVersion> pmvsInScope() {
-        return Stream.of(pmv, pmv); // same PMV twice simulates duplicate dependency declarations
+        return Stream.of(pmv, pmv);
       }
     };
     var impls = loader.load(ChatModelProvider.class);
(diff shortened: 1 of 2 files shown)


nhthinh-axonivy (changes requested):
The dedup fix itself (keying by `getClass().getName()` instead of relying on `distinct()`/identity) is correct, and `TestSpiLoader.load_noDuplicates_samePmvTwice` covers it well. That part looks good to merge.

There are two things I'd want addressed before this goes live:
### 1. Exception-narrowing regression in `loadRefs()` (`SpiLoader.java`)

```java
private static Set<String> loadRefs(Class<?> type, ClassLoader loader) {
  ...
  } catch (IOException ex) {   // was: catch (Exception ex)
  }
  return implRefs;
}

private static String read(InputStream service) {
  try (service) {
    return new String(service.readAllBytes(), StandardCharsets.UTF_8);
  } catch (IOException ex) {
    throw new RuntimeException(...);  // IOException gets wrapped as RuntimeException
  }
}
```

`read()` wraps its `IOException` in a plain `RuntimeException`. The narrowed `catch (IOException ex)` in `loadRefs()` no longer catches that wrapped exception, so a single unreadable/corrupt `META-INF/services` descriptor now crashes the entire SPI load, instead of returning the refs it had already collected, like before..

I'd suggest reverting this change (and the similar exception narrowing in `load(ClassLoader, String)`, which is lower risk but also outside this PR's scope) back to a broad `catch`. If exception narrowing is still desired, it would be better handled in a separate PR.

### 2. PR description doesn't match the implementation

The description says:

> Replace `distinct()` + `Collectors.toSet()` with `Collectors.toMap` keyed by `getClass().getName()`.

However, the implementation actually uses a manual `HashMap` + `putIfAbsent` loop.
The PR description should simply be updated to reflect what the code actually does.

nhthinh-axonivy (approved):
LGTM!

PR #335, opened 2026-07-08
Governance Part 4: Implement conversation details page

mnhnam-axonivy:
Adds a Conversations dialog to the AI Governance Center, allowing users to drill into per-case AI conversation history. The dialog displays a case → task → agent tree, with each agent showing its message timeline, token usage, tool calls, and response metadata.

image


ivy-rew:
I skip this one until we have the separated dashboard project; just add me again if you need feedback

PR #338, opened 2026-07-17
Add docusaurus build

ntloc-axonivy:
PR for introducing Docusaurus to Smart Workflow [AIAI-524](https://axon-ivy.atlassian.net/browse/AIAI-524)
Please refer to the ticket for more details.
This PR sets up Docusaurus as simply as possible using Docker Compose.

[AIAI-524]: https://axon-ivy.atlassian.net/browse/AIAI-524?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nhthinh-axonivy:
Spike

ivy-rew:
Cool, thanks for giving docusaurus a try @ntloc-axonivy . I can't wait to see a live demo. 🤩️

High level thoughts how to proceed:
- **versions**: are here by example duplicated into the filesystem. We should not require this, but simply upload versions of every release state to maven. And re-integrate existing versions via maven again. So the GIT/SCM will not be polluted with version management. But the build should provide them.
- **build**: the repo contains a lot of docusaurus resources which are required for building. We should craft a docker-image which contains all these good defaults. So that this repo needs minimal "docusaurus" config files to produce a doc. Still the image should allow overriding. Basically we're aiming for a stable central tooling as we do have it currently for the read-the-docs/sphynx stack. https://github.com/axonivy/build-container/tree/master/read-the-docs
- **hosting**: I see no attempt to actually publish the rendered documentation. I think we should evaluate this too ... just to be sure that there are no blockers when hosting this with free github.pages.
- **i18n**: I know that team-octopus wasn't super happy with weblate translation capabilities of .MD files. Also I don't know if .MDX is supporter. I think we should test how updates behave and whether there are unnecessarily content losses or merge issues if we set it up.

PR #340, opened 2026-07-21
refactor: adapt new process-extension-api

ivy-rew:
we deprecated most classes of the previous approach: and crafted a new jar.
the new artifact has the benefit of not even containing most of the new deprecated APIs. So things become clearer.


mnhnam-axonivy (approved):
Great!
please merge 👍

PR #341, opened 2026-07-21
Lovely emoji group icons 🤩️

ivy-rew:
improve attraction and recognizability by using emoji icons on group headers

tool-icons-bubble

emoji-headers

### DIFF
--- a/smart-workflow/src/com/axonivy/utils/smart/workflow/program/internal/AgentEditor.java
+++ b/smart-workflow/src/com/axonivy/utils/smart/workflow/program/internal/AgentEditor.java
@@ -23,21 +23,21 @@ import ch.ivyteam.ivy.process.program.ui.ProgramUiBuilder;
 public class AgentEditor {
 
   public void editor(ProgramUiBuilder ui) {
-    ui.group("Message")
+    ui.group("✉️ Message")
         .add(ui.label("User message:").create())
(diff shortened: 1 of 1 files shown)


smart-workflow/src/com/axonivy/utils/smart/workflow/program/internal/AgentEditor.java
```diff
@@ -23,21 +23,21 @@
 public class AgentEditor {
 
   public void editor(ProgramUiBuilder ui) {
-    ui.group("Message")
+    ui.group("✉️ Message")
```
ivy-rew, line 26: any strong feelings against using custom emojis on the inscription @ivy-lli ?
In my view it makes this rather complex inscription easier to use and groups are easier to recognize. But of course it makes the UI a bit more wild.

ivy-lli: I like that this works, however I'm not the biggest fan of emojis in such UI's. IMO as you said they make the UI very wild and they give a thing a bit an unprofessional look.
But I think you will find as many people like this as others. 
I wouldn't do this, but if you like it here 🤷‍♂️

mnhnam-axonivy (approved):
Look good to me 👍

PR #344, opened 2026-07-22
XIVY-18951 adapt new process element meta name api

ivy-rew:
We introduce an explicit API, so the ivyscript "hack" is no longer necessary.


mnhnam-axonivy (approved):
I love it! ❤️

PR #346, opened 2026-07-24
fix; run web-integration-test with engine props

ivy-rew:
fixes the test runtime setup ... but not the test itself.

### DIFF
--- a/smart-workflow-webtest/pom.xml
+++ b/smart-workflow-webtest/pom.xml
@@ -89,6 +89,7 @@
       <plugin>
         <groupId>com.axonivy.ivy.ci</groupId>
         <artifactId>project-build-plugin</artifactId>
+        <extensions>true</extensions>
         <version>${project.parent.version}</version>
       </plugin>
       <plugin>
--- a/smart-workflow-webtest/pom.xml
+++ b/smart-workflow-webtest/pom.xml
@@ -95,12 +95,8 @@
         <artifactId>maven-failsafe-plugin</artifactId>
         <configuration>
           <includes>
-            <include>**/*WebTest.java</include>
+            <include>**/*WebIT.java</include>
           </includes>
-          <systemPropertyVariables>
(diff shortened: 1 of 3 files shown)


smart-workflow-webtest/pom.xml
```diff
@@ -95,12 +95,8 @@
         <artifactId>maven-failsafe-plugin</artifactId>
         <configuration>
           <includes>
-            <include>**/*WebTest.java</include>
+            <include>**/*WebIT.java</include>
           </includes>
-          <systemPropertyVariables>
-            <test.engine.url>${test.engine.url}</test.engine.url>
-            <test.engine.app>${test.engine.app}</test.engine.app>
-          </systemPropertyVariables>
```
ivy-rew, line 103: properties should not be defined manually ... the project-build-plugin does this for us.

smart-workflow-webtest/pom.xml
```diff
@@ -95,12 +95,8 @@
         <artifactId>maven-failsafe-plugin</artifactId>
         <configuration>
           <includes>
-            <include>**/*WebTest.java</include>
```
ivy-rew, line 98: the previous runs failed; since still surefire was used to run the test: ... as the WebTest.java was also selected by surefire.
Basically we could explicitly exclude WebTEst from surefire. 
However, by using "IT" suffix we can work-around this issue. IT is auto-selected by failsafe. And ignored by surefire by default.

mnhnam-axonivy (approved):
thank you very much!

mnhnam-axonivy:
@ivy-rew let me fix the failed test then I will merge this branch

mnhnam-axonivy:
Hi @ivy-rew 

On VSCode everything working, but when run webtest, I got following error


I think it's related to recent change of the engine.

This is the action run:
https://github.com/axonivy-market/smart-workflow/actions/runs/30322822185/job/90162045652

So should I wait till we changed back to the snapshot version ?

PR #347, opened 2026-07-24
Switch to Jakarta Sprint.29 artifacts

ivy-rew:
let's go to a defined sprint release version:
- we are able to do an official first beta version release, that supports jakarta
- the ivy dev baseline is right now adopting breaking changes in IProcessModelVersion which we are depending on.

### DIFF
--- a/demo/smart-workflow-supplier-demo-test/pom.xml
+++ b/demo/smart-workflow-supplier-demo-test/pom.xml
@@ -79,6 +79,7 @@
       <plugin>
         <groupId>com.axonivy.ivy.ci</groupId>
         <artifactId>project-build-plugin</artifactId>
+        <version>${project.parent.version}</version>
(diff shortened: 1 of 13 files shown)


demo/smart-workflow-supplier-demo-test/pom.xml
```diff
@@ -79,6 +79,7 @@
       <plugin>
         <groupId>com.axonivy.ivy.ci</groupId>
         <artifactId>project-build-plugin</artifactId>
+        <version>${project.parent.version}</version>
```
ivy-rew, line 82: fyi @ivy-lmu : with this ref you don't even need to update the build-plugin ref if you move to the next parent.

mnhnam-axonivy (approved):
Let's go! 😄

PR #349, opened 2026-07-24
Back to leading edge dev-train

ivy-rew:
let's track the compatibility with ivy-core dev again ... after releasing the first jakarta beta


mnhnam-axonivy:
@ivy-rew just merge it once you've finished.

PR #351, opened 2026-07-28
Added inputVersion param as input to workflows

mnhnam-axonivy:
Allow overriding ivy engine version via workflow dispatch inputs

image

Adds an optional ivyVersion input to:
- dev
- ci
- e2e
- it
- release

workflows so the ivy engine permalink (e.g. dev, sprint, nightly) can be selected at dispatch time.

Defaults to `dev` to align with the upstream reusable workflow defaults.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,12 @@ on:
   schedule:
     - cron:  '21 21 * * *'
   workflow_dispatch:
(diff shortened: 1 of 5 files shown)


ivy-rew (approved):

ivy-rew:
very nice approach with parameterized builds :love_letter:

PR #357, opened 2026-07-30
XIYY-19081 multi-select input field for tools

ivy-rew:
use a widget with defined items for tools selection.
- migrate existing configs to new format: on inscription open and on execution (if needed)
- empty tool config will no longer evaluate to use all tools, rather we always expect a well-defined selection of tools

adapt https://github.com/axonivy/core/pull/13786
and https://github.com/axonivy/core/pull/13532

tool-selector

### DIFF
--- a/smart-workflow/src/com/axonivy/utils/smart/workflow/program/internal/AgentCallExecutor.java
+++ b/smart-workflow/src/com/axonivy/utils/smart/workflow/program/internal/AgentCallExecutor.java
@@ -1,5 +1,7 @@
 package com.axonivy.utils.smart.workflow.program.internal;
 
+import static com.axonivy.utils.smart.workflow.model.spi.ChatModelProvider.ModelOptions.options;
+
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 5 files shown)


smart-workflow/src/com/axonivy/utils/smart/workflow/program/internal/AgentCallExecutor.java
```diff
@@ -73,7 +74,7 @@ public void execute() {
     var agentBuilder = AiServices.builder(agentType);
     var memory = configureMemory(agentBuilder);
     var human = configureHumanInTheLoop(memory, agentBuilder);
-    var toolFilter = executeListOfStrings(Conf.TOOLS).orElse(null);
+    var toolFilter = context.config().getList(Conf.TOOLS);
```
ivy-rew, line 77: accepted side-effect: tools can no longer be scripted. Values like "in.myTools" can no longer be evaluated, if we maintain a widget with a defined list of items.
I think this is no big deal, if customers really need dynamic scripted agent definitions, we or the customer can easily fork our AgentCallExecutor and re-introduce a variant that supports scripting just like it was until now.

---

This effect appears here for the tools by example: I think the same multi-select component can be used for in-/output guardrails+model-provider selection. 
Do you think we need scripting in any of these @mnhnam-axonivy

smart-workflow/src/com/axonivy/utils/smart/workflow/program/internal/AgentEditor.java
```diff
@@ -67,42 +68,40 @@ public void editor(ProgramUiBuilder ui) {
...
       var pmv = SpiProject.getSmartWorkflowPmv();
       return new SpiLoader(pmv).load(SmartWorkflowToolsProvider.class).stream()
           .flatMap(provider -> {
             var tools = provider.getTools();
             return tools == null ? Stream.empty() : tools.stream();
           })
-          .map(SmartWorkflowTool::name)
-          .distinct()
-          .toList();
+          .map(t -> SelectItem.of(t.name(), t.name(), "☕️", t.description()));
```
ivy-rew, line 102: by now the tool icons are static. but we can easily extend it in a follow up to be dynamically defined by the tool provider. 
like in the prototype below icons can bring a lot of fun into these:

Image

ivy-rew:
chat discussion: there's an integration problem currently in the vscdesigner (class-loading) ... but we integrate this as anyway; since it should not relate to the problem at hand.

PR #360, opened 2026-07-30
Migrate all smart-workflow projects to new extension.api

ivy-rew:
without these dependencies the validation fails. seems to be a false-positive though, since the provided dep from the referenced 'smart-workflow:iar' should be considered.

~-dev-market-smart-workflow-2_001

[mvn.log.txt](https://github.com/user-attachments/files/30544993/mvn.log.txt)

### DIFF
--- a/demo/smart-workflow-supplier-demo/pom.xml
+++ b/demo/smart-workflow-supplier-demo/pom.xml
@@ -17,8 +17,8 @@
       <artifactId>ivy-api</artifactId>
     </dependency>
     <dependency>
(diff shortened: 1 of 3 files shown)


ivy-rew:
fyi @ivy-edp 

I think this is a bug of the validation mojo. The process-extension-api is already registered in the common IAR dependency of these demo/test projects -> therefore a re-declaration should not be necessary https://github.com/axonivy-market/smart-workflow/blob/9bf9d1d2dc8a842700df34443210784fe7ac7408/smart-workflow/pom.xml#L24

PR #361, opened 2026-07-30
Inherited dependencies from smart-workflow

ivy-rew:
Do not enforce consumers of smart-workflow to re-declare the program-api.


smart-workflow/pom.xml
```diff
@@ -23,6 +23,7 @@
     <dependency>
       <groupId>com.axonivy.ivy.api.extension</groupId>
       <artifactId>ivy-process-extension-api</artifactId>
+      <scope>compile</scope>
     </dependency>
```
ivy-rew, line 27: seems to be necessary to compile/test/validate. consequence is that the smart-workflow:iar will contain these libs. is there a better solution that I'm missing @ivy-cst ?

Image

PR #369, opened 2026-08-10
AIAI-585 Application scope circuit breaker

mnhnam-axonivy:
### Application-wide Circuit Breaker

- A kill switch that stops all AI agent calls, independent of any per-agent guardrail configuration.
Implemented as a regular guardrail, so it reuses the existing guardrail pipeline instead of introducing a separate check path. 

- This gives the circuit breaker several advantages:
  + Easy to record: It uses the same listener mechanism as other guardrails.
  + Allow fallback: Users can define an alternative solution when an agent is stopped, meaning the overall system can continue working without AI.

image

### Interesting points

Currently, I use the Ivy variable `AI.CircuitBreaker.StopAll`. Admins can:

- Set it to false: all agents are stopped and cannot run, even when a user starts a new process.
- Set it to true: all agents are allowed to start working again.

@ivy-rew, do you think using an Ivy variable is a suitable approach for this?

Introduce a new `alwaysOn` attribute for guardrails. A guardrail with this attribute will always run and will not appear in the guardrail list, so developers cannot add or remove it.

### Demo

Process `ApplicationScope.p.json` has 2 process starts:

- `runApplicationAgentsDemo`: to start a time consuming agent process
- `stopApplicationAgents`: set the variable `StopAll` to true, so all running agents will stop


smart-workflow/src/com/axonivy/utils/smart/workflow/permission/SmartWorkflowPermission.java
```diff
@@ -0,0 +1,16 @@
+package com.axonivy.utils.smart.workflow.permission;
+
+public enum SmartWorkflowPermission {
+    
+    ADMIN("AiAdmin");
```
ivy-rew, line 5: shouldn't we add this role to smart-workflow, to make it known?

smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/entity/SmartWorkflowGuardrail.java
```diff
@@ -10,4 +10,8 @@ default GuardrailResult evaluate(String message, String invocationId) {
   default String name() {
     return getClass().getSimpleName();
   }
+
+  default boolean alwaysOn() {
+    return false;
+  }
```
ivy-rew, line 16: I think we should make this switch more internal. Now it could motivate users to enable their favorite guardrail by default, just because everyone likes his own invention. 

I'd rather introduce an additional interface (my extends SmartWorkflowGuardrail) which we use only to flag "alwaysOn" guardrails. I'd keep this an interna, not motivating third-parties to use it until we see other good use-cases emerging.

smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/circuitbreaker/internal/VariableCircuitBreakerSignal.java
```diff
@@ -0,0 +1,39 @@
...
+    } catch (RuntimeException ex) {
+      return Optional.empty();
+    }
+  }
+
+  private static String buildReason() {
+    try {
+      String user = Ivy.session().getSessionUserName();
+      String time = LocalDateTime.now().format(DatePatternUtils.dateTimeFormatter());
+      return "All AI agent activities are stopped by " + user + " at " + time;
```
ivy-rew, line 34: is this accurate? or where is that state, of the user that initially disabled the execution via circuit breaker?

smart-workflow/config/variables.yaml
```diff
@@ -3,11 +3,18 @@ Variables:
   AI:
     # [enum: OpenAI, AzureOpenAI, Gemini, xAI, Anthropic, Ollama]
     DefaultProvider: ""
+    # Emergency circuit breaker. Blocks AI agent calls at the guardrail layer.
+    # A tool loop already in flight finishes before the block takes effect.
+    CircuitBreaker:
+      # "true" stops ALL AI agent calls in this application. Case-insensitive; also accepts yes/on/1.
+      StopAll: ""
```
ivy-rew, line 10: I'd not encourage the other "trueish" values. Just document true/false.

smart-workflow/config/variables.yaml
```diff
@@ -3,11 +3,18 @@ Variables:
...
     # Guardrails designed to ensure that AI operate safely ethical, and legal
     Guardrails:
       # Default input guardrails. Separated by comma. Available: PromptInjectionInputGuardrail, AiPromptInjectionInputGuardrail, PiiMaskingGuardrail
+      # (the always-on circuit breaker guardrail runs in addition to these)
```
ivy-rew, line 14: I think the fact that we implemented the CircuitBreaker as a guardrail is an implementation detail. 
So I'd not document it here ... I think it will make the user think about something he doesn't need to care about.

smart-workflow/config/variables.yaml
```diff
@@ -3,11 +3,18 @@ Variables:
...
       DefaultInput: ""
       # Default output guardrails. Separated by comma. Available: SensitiveDataOutputGuardrail, PiiMaskingGuardrail
+      # (the always-on circuit breaker guardrail runs in addition to these)
```
ivy-rew, line 17: omit

smart-workflow-demo/process/Patterns/CircuitBreaker/ApplicationScope.p.json
```diff
@@ -0,0 +1,285 @@
...
+        "Stop all agents in ",
+        "application"
+      ],
+      "config" : {
+        "output" : {
+          "code" : [
+            "import com.axonivy.utils.smart.workflow.utils.IvyVar;",
+            "import com.axonivy.utils.smart.workflow.guardrails.circuitbreaker.CircuitBreakerSignal;",
+            "",
+            "IvyVar.set(CircuitBreakerSignal.STOP_ALL_VARIABLE, \"true\");"
```
ivy-rew, line 36: I think Ivy.var().set() is better here. Adding another utility for a very well known API makes the user guess what else is in there.

smart-workflow/src/com/axonivy/utils/smart/workflow/permission/PermissionValidator.java
```diff
@@ -0,0 +1,12 @@
+package com.axonivy.utils.smart.workflow.permission;
+
+import ch.ivyteam.ivy.environment.Ivy;
+import ch.ivyteam.ivy.security.IRole;
+
+public final class PermissionValidator {
+    public static boolean isAiAdmin() {
```
ivy-rew, line 7: cool idea to introduce a specific permission, but we have no means to enforce it, right?
anyone can write code that sets the variable. currently we also didn't use it in the demo, afaik.
how will it be used?

smart-workflow/config/variables.yaml
ivy-rew, line 10: Maybe `AI.CircuitBreaker.Enabled=true/false` would be more intuitive.

smart-workflow/config/variables.yaml
```diff
@@ -3,11 +3,18 @@ Variables:
   AI:
     # [enum: OpenAI, AzureOpenAI, Gemini, xAI, Anthropic, Ollama]
     DefaultProvider: ""
+    # Emergency circuit breaker. Blocks AI agent calls at the guardrail layer.
+    # A tool loop already in flight finishes before the block takes effect.
+    CircuitBreaker:
```
ivy-rew, line 8: I think the variable is a good starting point for LTS14. On the long run I think we need to be able to disable only a selection of ongoing calls: only those of my sessions, or only one that is running for minutes without results.

smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/circuitbreaker/CircuitBreakerSignal.java
```diff
@@ -0,0 +1,16 @@
+package com.axonivy.utils.smart.workflow.guardrails.circuitbreaker;
+
+import java.util.Optional;
+
+import com.axonivy.utils.smart.workflow.guardrails.circuitbreaker.internal.VariableCircuitBreakerSignal;
+
+public interface CircuitBreakerSignal {
+
+  String STOP_ALL_VARIABLE = "AI.CircuitBreaker.StopAll";
```
ivy-rew, line 9: I think we better keep the string an interna, but provide a static method here to fire the signal.
In here we check the permissions of the callers, and set the variable.
This gives us more flexibility to change and makes usage simpler (less ivy-script imports and no string-boolean passing for the user)

smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/circuitbreaker/CircuitBreakerGuardrail.java
```diff
@@ -0,0 +1,30 @@
+package com.axonivy.utils.smart.workflow.guardrails.circuitbreaker;
+
+import com.axonivy.utils.smart.workflow.guardrails.entity.GuardrailResult;
+import com.axonivy.utils.smart.workflow.guardrails.entity.SmartWorkflowInputGuardrail;
+import com.axonivy.utils.smart.workflow.guardrails.entity.SmartWorkflowOutputGuardrail;
+
+public class CircuitBreakerGuardrail implements SmartWorkflowInputGuardrail, SmartWorkflowOutputGuardrail {
```
ivy-rew, line 7: I like the fact, that we can fire a clear signal with a bpmError to process implementors, so implementing this circuit breaker as guardrail is a good idea.

ivy-rew: Have you explored how other harnesses implement a circuit-breaker? I think the ongoing request could still take a lot of processing time ... if the called LLM is producing a lot of output-tokens. And we can't stop it until we get the final message with the current approach. 

Shouldn't we look for a solution to just close the ongoing http-request?

ivy-rew (commented):
Cool first solution. A raised a few discussion points below....

mnhnam-axonivy:
Hi @ivy-rew 
I introduced two internal guardrail types: `SmartWorkflowInternalInputGuardrail` and `SmartWorkflowInternalOutputGuardrail`.
These are input/output guardrails with the following characteristics:
  - always run
  - cannot be removed by developers

And the execute order is:

internal input guardrails
→ internal guardrails
→ LLM calls
-> internal output guardrails
→ output guardrails

With this approach, we no longer need to expose the `alwaysOn` attribute, and it also provides a clear path for adding future internal guardrails (if any 😄).

Details of code changes
https://github.com/axonivy-market/smart-workflow/pull/369/changes/a45531e8fbea760e04e414539edb01299b99a131

smart-workflow/src/com/axonivy/utils/smart/workflow/permission/PermissionValidator.java
ivy-rew: I think we better remove the perm, since we can't really enforce it.
The core would need a variables permission concept to actually verify it.

smart-workflow/src/com/axonivy/utils/smart/workflow/permission/SmartWorkflowPermission.java
mnhnam-axonivy: I removed permission. Commit: https://github.com/axonivy-market/smart-workflow/pull/369/changes/6405a7e993bea9c6a041805df5981b72bae9dd08

smart-workflow/src/com/axonivy/utils/smart/workflow/permission/PermissionValidator.java
mnhnam-axonivy: I removed permission
https://github.com/axonivy-market/smart-workflow/pull/369/changes/6405a7e993bea9c6a041805df5981b72bae9dd08

smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/circuitbreaker/CircuitBreakerGuardrail.java
mnhnam-axonivy: I will research it in another story together with how to stop a dedicated agent
https://axon-ivy.atlassian.net/browse/AIAI-603

smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/circuitbreaker/internal/VariableCircuitBreakerSignal.java
mnhnam-axonivy: dropped in https://github.com/axonivy-market/smart-workflow/pull/369/changes/929955db8f7d9557e73478121eb9db78044c3694

smart-workflow/config/variables.yaml
mnhnam-axonivy: updated in https://github.com/axonivy-market/smart-workflow/pull/369/changes/9a30df4d6dfc02bdb01a49605df67fd6689766ec

smart-workflow/config/variables.yaml
mnhnam-axonivy: Will be investigate and implement in https://axon-ivy.atlassian.net/browse/AIAI-603

smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/circuitbreaker/CircuitBreakerSignal.java
mnhnam-axonivy: Fixed in https://github.com/axonivy-market/smart-workflow/pull/369/changes/929955db8f7d9557e73478121eb9db78044c3694

nhthinh-axonivy:
@ivy-rew I'm verifying AIAI-585. Two quick things before I approve:

1. Confirming my read of the thread: permission removal (can’t be reliably enforced) and in-flight cancellation/per-session stop (out of scope) are both settled as it is?

2. Your “omit” comment on `variables.yaml` is still open, and the code still says “guardrail layer” in both the variable comment and today’s new doc. Is that a must-fix before this is considered done, or is it fine to leave as it is?

Also noticed there’s no approval yet on the latest commits (rename + doc landed after your last review). Could you take a recheck and approve when you get a chance? I’ll hold verify this until then.

ivy-rew:
> @ivy-rew I'm verifying AIAI-585. Two quick things before I approve:
> 
>     1. Confirming my read of the thread: permission removal (can’t be reliably enforced) and in-flight cancellation/per-session stop (out of scope) are both settled as it is?
> 
>     2. Your “omit” comment on `variables.yaml` is still open, and the code still says “guardrail layer” in both the variable comment and today’s new doc. Is that a must-fix before this is considered done, or is it fine to leave as it is?
> 
> 
> Also noticed there’s no approval yet on the latest commits (rename + doc landed after your last review). Could you take a recheck and approve when you get a chance? I’ll hold verify this until then.

Thanks @nhthinh-axonivy 

1. Yes permissions can't be enforced, so I think having a first version that does no extra checks here is fine. At least it doesn't give the user a false assurance.
2. No strong feelings from my side on this guardrail comment. I'd remove it is an implementation detail, which we may change in a sub-sequent iteration. 

I browsed the docs real quick, and I think its good.

smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/entity/GuardrailInspectionText.java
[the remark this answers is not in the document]
mnhnam-axonivy: sorry my mistake when merge code

mnhnam-axonivy:
@nhthinh-axonivy , I removed unnecessary comments in variable file and remove the duplicated class
commit: https://github.com/axonivy-market/smart-workflow/pull/369/changes/6ca17a6331d4eef990d8a9a08440f5921a678a9d

nhthinh-axonivy (approved):
LGTM!

PR #370, opened 2026-08-10
AIAI-598: Guardrail throws error when user message has file content

mnhnam-axonivy:
When run an agent with type different from text, for example:

```
Extract all text content from this invoice image: <%=in.invoiceStream%>
```

An error will throw when run through input guardrails

image

---

### Solution

replace the content of different types with placehoder. For example:

```
Extract all text content from this invoice image: <file: IMAGE>
```

### DIFF
--- a/smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/entity/GuardrailInspectionText.java
+++ b/smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/entity/GuardrailInspectionText.java
@@ -1,0 +1,49 @@
+package com.axonivy.utils.smart.workflow.guardrails.entity;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import dev.langchain4j.data.message.Content;
+import static dev.langchain4j.data.message.ContentType.AUDIO;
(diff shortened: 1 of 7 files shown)


smart-workflow/src/com/axonivy/utils/smart/workflow/guardrails/entity/GuardrailInspectionText.java
```diff
@@ -0,0 +1,49 @@
+package com.axonivy.utils.smart.workflow.guardrails.entity;
```
ivy-rew: I think this is a helper for the 'InputGuardrailAdapter' and 'Listener'. So I'd also place this class there in the "adapter" package. To keep it a pure interna.

ivy-rew: The 'entity' package today is used for the public interfaces. I think we will need to shift these too. As I think these will be consumed and implemented by third-parties. So keeping them in the root of guardrails seems more appropriate to me -> `com.axonivy.utils.smart.workflow.guardrails`  ... but that's another discussion and should be cleaned in another PR:

ivy-rew (approved):
Thanks a lot for the fix. 🤝

PR #380, opened 2026-08-21
Convert file upload to primefaces 15

mnhnam-axonivy:
In the supplier demo project, we got error related to `sizeLimit`

Screenshot 2026-08-21 103045

Root cause (verified against primefaces-15.0.17-jakarta.jar's taglib):

- PrimeFaces 15 removed sizeLimit, allowTypes, and fileLimit from p:fileUpload and moved them to a new nested validator <p:validateFile> (validator id primefaces.File).
- dragDropSupport was also renamed to dragDrop.

See https://primefaces.github.io/primefaces/15_0_0/#/../migrationguide/15_0_0?id=fileupload


ivy-rew (approved):
Thanks a lot 🤝

PR #381, opened 2026-08-21
Disable xhtml validator of project-build-plugin

mnhnam-axonivy:
some projects show XHTML warnings but currently I have no idea why... code seem valid and run correctly.
Therefore I want to temporary disable them for now.

Screenshot 2026-08-21 135233


ivy-rew:
Thanks @mnhnam-axonivy . We have similar issues in the core and are still on the way of improving.
FYI @ivy-edp , there are market projects that monitor warnings :wink:

PR #385, opened 2026-08-24
Agent executor without dynamic interface building

ivy-rew:
getting ready to leave beta-phase:
- drop runtime class creation just to persuade langchain4j services to adopt our return type.

---

seems like the AIContext already provides the required means to override the return type dynamically. 🎯️


smart-workflow/pom.xml
```diff
@@ -95,11 +95,6 @@
       <artifactId>ivy-java-api</artifactId>
       <scope>provided</scope>
     </dependency>
-    <dependency>
-      <groupId>org.ow2.asm</groupId>
-      <artifactId>asm</artifactId>
-      <version>9.10.1</version>
-    </dependency>
```
ivy-rew: love to see this adventurous runtime class creation leaving 👋 🥲 🗑️

mnhnam-axonivy (approved):
Great works! 👍

PR #387, opened 2026-08-25
Initialize document workflow

mnhnam-axonivy:
### When the workflow runs

Documentation workflow that build the Docusaurus project `smart-workflow-document` and deploy to Github Page. This workflow runs:

- When a pull request is created/updated → build only
- When pushed to master or manually triggered → build + deploy

and when one of these paths changes:

```
smart-workflow-document/**
.github/workflows/docs.yml
```
This avoids unnecessary documentation builds when unrelated files are changed.

### Permissions
`contents: read`: Allows the workflow to read the repository.
`pages: write`: Allows the workflow to deploy the generated site to GitHub Pages.
`id-token: write`: Used by GitHub Pages deployment for authentication.

### Artifacts

For pull requests, the generated documentation build is uploaded as a GitHub Actions artifact and is kept for 14 days.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,80 @@
+name: Docs-Build
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'smart-workflow-document/**'
+      - '.github/workflows/docs.yml'
+  pull_request:
+    paths:
+      - 'doc/**'
+      - 'smart-workflow-document/**'
+      - '.github/workflows/docs.yml'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+jobs:
+  build:
+    name: Build docs
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7
+
+      - name: Setup Node
+        uses: actions/setup-node@v7
+        with:
+          node-version: 22
+          cache: npm
+          cache-dependency-path: smart-workflow-document/package-lock.json
+
+      - name: Install dependencies
+        working-directory: smart-workflow-document
+        run: npm ci
+
+      - name: Type check
+        working-directory: smart-workflow-document
+        run: npm run typecheck
+
+      - name: Build site
+        working-directory: smart-workflow-document
+        run: npm run build
+
+      - name: Upload Pages artifact
+        if: github.event_name != 'pull_request'
+        uses: actions/upload-pages-artifact@v5
+        with:
+          path: smart-workflow-document/build
+
+      - name: Upload PR preview build
+        if: github.event_name == 'pull_request'
+        uses: actions/upload-artifact@v7
+        with:
+          name: docs-preview-pr-${{ github.event.pull_request.number }}
+          path: smart-workflow-document/build
+          retention-days: 14
+
+  deploy:
+    name: Deploy to GitHub Pages
+    needs: build
+    if: github.event_name != 'pull_request'
+    runs-on: ubuntu-latest
+    concurrency:
+      group: pages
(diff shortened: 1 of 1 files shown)


mnhnam-axonivy:
Hi @ivy-rew 

This is the workflow for the upcoming document build using Docusaurus.
The build is currently failing because it requires the source from the `smart-workflow-document` project.

Please take a look at the content generated by this build and give me your feedback. I’d like to merge it into `master` so I can try deploying the documentation to GitHub Pages.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,80 @@
...
+        working-directory: smart-workflow-document
+        run: npm ci
+
+      - name: Type check
+        working-directory: smart-workflow-document
+        run: npm run typecheck
+
+      - name: Build site
+        working-directory: smart-workflow-document
+        run: npm run build
```
ivy-rew: ok for a first try-out. But I think for the final version we don't want docusaurus NPM infrastructure in here. But rather just run it from a docker-container. Just like we do currently for readthedocs

ivy-rew (approved):

.github/workflows/docs.yml
```diff
@@ -0,0 +1,80 @@
...
+        uses: actions/upload-pages-artifact@v5
+        with:
+          path: smart-workflow-document/build
+
+      - name: Upload PR preview build
+        if: github.event_name == 'pull_request'
+        uses: actions/upload-artifact@v7
+        with:
+          name: docs-preview-pr-${{ github.event.pull_request.number }}
+          path: smart-workflow-document/build
```
ivy-rew: I think I'd just use `docs/user` ..

.github/workflows/docs.yml
mnhnam-axonivy: Does it mean we could have a docker image like `axonivy/build-container:read-the-docs-2` in near future?

Nice! 😄

ivy-rew: Yes, I think we should go there. There are many benefits:
- no need to deal with NPM in our java repos
- faster execution as we just use a shared image
- ease of use for everyone :smile:

.github/workflows/docs.yml
mnhnam-axonivy: Sorry, I don't understand, do you mean rename `smart-workflow-document` to `docs`?

or keep the `smart-workflow-document` as is, just move the content to `docs/user`  as below

```
docs/
  user/                     ← content (was doc/)
smart-workflow-document/    ← unchanged
  docusaurus.config.ts      → docs.path: '../docs/user'
  versioned_docs/  i18n/  src/  static/  package.json
```

currently it's the folder we store the Docusaurus project:

image

ivy-rew: I think contents in /doc today could already help the end-user. But maybe there are also technical things (testing), which are only important for contributors of the repo. So I'd separate them:
/doc/user <- end user doc (most of the currenct /doc contents)
/doc/dev <- developer docs (markdown only ... no docusaurus processing)

The smart-workflow-document looks like the prototype of Loc. I think I would not add this to this repository in this state. This should be something that is produced at build time and is not under version control. 
- versioned_docs ... should be injected via maven or from older GIT branches
- NPM infra and static contents should live in a docker container (maintained outside of this repo)
- i18n may live in /doc/i18n/... 
- large images used in our docs, should not be under version control (makes GIT slow). We need a build capturing them and sharing it as maven artifact. Which we unpack at build time into our docs.
- a build should simply read our /doc/user and craft a Docusaurus page out of it and upload it as github page.

mnhnam-axonivy:
I will close this PR
The implementation of workflow + docker image + screenshot build will be discuss and implement in story
https://axon-ivy.atlassian.net/browse/AIAI-607

PR #394, opened 2026-08-26
XIVY 19326 Optional Dependency installation

ivy-tgr:


ivy-rew (approved):

smart-workflow-product/product.json
```diff
@@ -52,42 +52,48 @@
...
+            "requiredOneOf": "provider"
           },
           {
             "groupId": "com.axonivy.utils.ai",
             "artifactId": "smart-workflow-gemini",
             "version": "${version}",
             "type": "iar",
-            "optional": true
+            "optional": true,
+            "requiredOneOf": "provider"
```
ivy-rew: Sorry, but after a night of sleep, I think we should rename this property again 😔 . I think it sounds like a typo. `requireOneOf` would be more natural to me. @ivy-tgr

